PR #6, opened 2018-04-11
Create endpoints for compute resource.

GustavoDinizMonteiro:
PR associated with task [#458](https://git.lsd.ufcg.edu.br/chicog/fogbow/issues/458).

### DIFF
--- a/src/main/java/org/fogbowcloud/manager/core/controllers/ComputeOrdersController.java
+++ b/src/main/java/org/fogbowcloud/manager/core/controllers/ComputeOrdersController.java
@@ -1,0 +1,38 @@
+package org.fogbowcloud.manager.core.controllers;
+
+import java.util.List;
+
+import org.fogbowcloud.manager.core.models.orders.ComputeOrder;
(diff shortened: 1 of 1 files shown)


src/main/java/org/fogbowcloud/manager/core/controllers/ComputeOrdersController.java
```diff
@@ -0,0 +1,38 @@
...
+import org.springframework.web.bind.annotation.RequestMapping;
+import org.springframework.web.bind.annotation.RequestMethod;
+import org.springframework.web.bind.annotation.RestController;
+
+@CrossOrigin
+@RestController
+@RequestMapping(value = "compute")
+public class ComputeOrdersController {
+	@RequestMapping(method = RequestMethod.POST)
+	public ResponseEntity<ComputeOrder> createCompute(@RequestBody ComputeOrder ComputeOrder) {
```
paulofelipefeitosa: The HTTP POST return should not be a ComputeOrder, if think that should be an OK or ERROR status (String Representation).

PR #9, opened 2018-04-12
Create interface between REST and the Controller called ApplicationController

wassoares:
Task: #517.


GustavoDinizMonteiro:
No needed anymore.

PR #12, opened 2018-04-16
Compute persistence #516

GustavoYamaguchi:
- Making models persistent.
- Creating Datastore Interface (provides contract, this ensure our database to be easily changeable).
- Attaching JPA to Datastore implementation.

### DIFF
--- a/src/main/java/org/fogbowcloud/manager/core/models/orders/Order.java
+++ b/src/main/java/org/fogbowcloud/manager/core/models/orders/Order.java
@@ -12,9 +12,9 @@ import org.fogbowcloud.manager.core.models.token.Token;
 @Inheritance(strategy = InheritanceType.SINGLE_TABLE)
 @Table(name = "tb_order")
 @JsonTypeInfo(use = JsonTypeInfo.Id.NAME, include = JsonTypeInfo.As.EXTERNAL_PROPERTY, property = "type")
-@JsonSubTypes({ @JsonSubTypes.Type(value = ComputeOrder.class, name = "COMPUTE"),
-		@JsonSubTypes.Type(value = NetworkOrder.class, name = "NETWORK"),
-		@JsonSubTypes.Type(value = StorageOrder.class, name = "STORAGE")})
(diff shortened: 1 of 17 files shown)


src/main/java/org/fogbowcloud/manager/core/models/orders/instances/OrderInstance.java
```diff
@@ -1,14 +1,56 @@
...
     public String getId() {
         return id;
     }
 
     public void setId(String id) {
         this.id = id;
     }
+
+	@Override
+	public int hashCode() {
```
paulofelipefeitosa: hashCode and equals methods could be implemented at the Order superclass for code reuse.

src/main/java/org/fogbowcloud/manager/core/rest/ComputeOrdersController.java
```diff
@@ -16,9 +19,13 @@
...
 @RequestMapping(value = "compute")
 public class ComputeOrdersController {
+	@Autowired
+	DatastoreServiceImpl orderService;
+
 	@RequestMapping(method = RequestMethod.POST)
-	public ResponseEntity<String> createCompute(@RequestBody ComputeOrder ComputeOrder) {
-		return new ResponseEntity<String>(HttpStatus.OK);
+	public ResponseEntity<Order> createCompute(@RequestBody ComputeOrder computeOrder) {
+		System.out.println(computeOrder);
```
paulofelipefeitosa: This should be here?

src/main/java/org/fogbowcloud/manager/core/models/orders/Order.java
```diff
@@ -117,6 +166,8 @@ public boolean equals(Object obj) {
 			if (other.id != null)
 				return false;
 		} else if (!id.equals(other.id))
+			return false;
```
paulofelipefeitosa: I think that the equals method only in the OrderId could be enough :).

PR #14, opened 2018-04-23
Double linked list cont #528

GustavoYamaguchi:


src/main/java/org/fogbowcloud/manager/core/models/linkedlist/Node.java
```diff
@@ -0,0 +1,42 @@
...
+
+import org.fogbowcloud.manager.core.models.orders.Order;
+
+public class Node {
+
+	private Node next;
+	private Node previous;
+	private Order order;
+
+	public Node(){ }
```
fubica: Do we need this constructor? This will create a Node with an Order null. We assume no element in the list has an Order that is null.

src/main/java/org/fogbowcloud/manager/core/models/linkedlist/SynchronizedDoubleLinkedList.java
```diff
@@ -0,0 +1,108 @@
+package org.fogbowcloud.manager.core.models.linkedlist;
+
+import org.fogbowcloud.manager.core.models.orders.Order;
+
+public class SynchronizedDoubleLinkedList implements ChainedList {
```
fubica: Change Double by Doubly in the name of the class.

src/main/java/org/fogbowcloud/manager/core/models/linkedlist/SynchronizedDoubleLinkedList.java
```diff
@@ -0,0 +1,108 @@
...
+	}
+
+	protected synchronized Node getTail() {
+		return this.tail;
+	}
+
+	@Override
+	public synchronized void addItem(Order order) {
+		if (order == null) {
+			throw new IllegalArgumentException("Order cannot be null.");
```
fubica: Change message for: "Attempting to add a null order."

src/main/java/org/fogbowcloud/manager/core/models/linkedlist/SynchronizedDoubleLinkedList.java
```diff
@@ -0,0 +1,108 @@
...
+		}
+		if (this.head == null) {
+			Node firstNode = new Node(null, order, null);
+			this.tail = this.head = this.current = firstNode;
+		} else {
+			Node newItem = new Node(this.tail, order, null);
+			this.tail.setNext(newItem);
+			this.tail = newItem;
+		}
+	}
```
fubica: We need to check if current was in the end of the list (current == null). In this case, current should point to the newItem just added. Simply add:

    if (this.current == null) this.current = this.tail;

Add a comment to explain this, as it is done in the removeItem() method.

src/main/java/org/fogbowcloud/manager/core/models/linkedlist/SynchronizedDoubleLinkedList.java
```diff
@@ -0,0 +1,108 @@
...
+
+	@Override
+	public synchronized Order getNext() {
+		if (this.current == null) {
+			return null;
+		}
+		Order orderCurrent = this.current.getOrder();
+		Node nextNode = this.current.getNext();
+		this.current = nextNode;
```
fubica: Why not simply "this.current = this.current.getNext();"? We don't need to create the temporary Node (nextNode).

src/main/java/org/fogbowcloud/manager/core/models/linkedlist/SynchronizedDoubleLinkedList.java
```diff
@@ -0,0 +1,108 @@
...
+	 * order, and point current to the next one.
+	 *
+	 * @param order
+	 * @return True if it was removed from the list.
+	 * 		   False, if another thread removed this order from the list and the order couldn't be find.
+	 */
+	@Override
+	public synchronized boolean removeItem(Order order) {
+		if (order == null) {
+			throw new IllegalArgumentException("Order cannot be null.");
```
fubica: Change message for: "Attempting to remove a null order."

src/main/java/org/fogbowcloud/manager/core/models/linkedlist/SynchronizedDoubleLinkedList.java
```diff
@@ -0,0 +1,108 @@
...
+		return true;
+	}
+
+	/**
+	 * @param order - Never null
+	 */
+	protected synchronized Node findNodeToRemove(Order order) {
+		Node currentNode = this.head;
+		while (currentNode != null) {
+			if (order.equals(currentNode.getOrder())) {
```
fubica: We compare the reference to the order, an not the other itself. Instead of equals, use ==.

PR #18, opened 2018-04-25
Removing DB dependency from project

GustavoYamaguchi:


src/main/resources/application.properties
```diff
@@ -1,8 +1,14 @@
+#DB-Free:
+spring.autoconfigure.exclude=org.springframework.boot.autoconfigure.orm.jpa.HibernateJpaAutoConfiguration,\
+    org.springframework.boot.autoconfigure.jdbc.DataSourceAutoConfiguration,\
+    org.springframework.boot.autoconfigure.jdbc.DataSourceTransactionManagerAutoConfiguration,\
+    org.springframework.boot.autoconfigure.data.web.SpringDataWebAutoConfiguration
+
```
arn3tt: Is this duplicated? See the EnableAutoConfiguration annotation on FogbowManagerApplication.

GustavoYamaguchi: Yes it is, could you review this again?

PR #19, opened 2018-04-25
Implement SharedDataStructures

italo-batista:


src/main/java/org/fogbowcloud/manager/core/datastructures/SharedDataStructures.java
```diff
@@ -0,0 +1,72 @@
+package org.fogbowcloud.manager.core.datastructures;
+
+import org.fogbowcloud.manager.core.models.linkedList.SynchronizedDoublyLinkedList;
+import org.fogbowcloud.manager.core.models.orders.Order;
+
+import java.util.Map;
+import java.util.concurrent.ConcurrentHashMap;
+
+public class DataSharedStructure {
```
fubica: Use SharedDataStructures.

src/main/java/org/fogbowcloud/manager/core/datastructures/SharedDataStructures.java
```diff
@@ -0,0 +1,72 @@
...
+            }
+            return instance ;
+        }
+    }
+
+    public Map<String, Order> getActiveOrdersMap() {
+        return activeOrdersMap;
+    }
+
+    public SynchronizedDoublyLinkedList getOpenOrders() {
```
fubica: Add Lists to the end of all get<state>Orders() calls.

src/test/java/org/fogbowcloud/manager/core/datastructures/SharedDataStructureTest.java
```diff
@@ -0,0 +1,50 @@
...
+    public void initialize() {
+        this.instanceOne = DataSharedStructure.getInstance();
+        this.instanceTwo = DataSharedStructure.getInstance();
+    }
+
+    @Test
+    public void testGetSameListReference() {
+
+        SynchronizedDoublyLinkedList listFromInstanceOne = instanceOne.getOpenOrders();
+        SynchronizedDoublyLinkedList listFromInstanceTwo = instanceTwo.getOpenOrders();
```
fubica: Add an assertEquals here.

italo-batista:
@thiagomanel @fubica 
What about SharedOrdersStructures, since is a class specialized in provide data structures, related to Orders, that must be shared through all the code?

fubica:
I see Thiago's point, and he is right. I suggest use SharedOrderHolders.

PR #20, opened 2018-04-26
Create OrderStateTransitioner

arn3tt:


src/main/java/org/fogbowcloud/manager/core/datastructures/OrderStateTransitioner.java
```diff
@@ -0,0 +1,73 @@
...
+import org.fogbowcloud.manager.core.models.orders.OrderState;
+
+public class OrderStateTransitioner {
+
+    private static final Logger LOGGER = Logger.getLogger(OrderStateTransitioner.class);
+
+    /**
+     * Updates the state of the given order. This method is not
+     * thread-safe. The caller must assure that only one
+     * thread can call it at the same time.
```
fubica: Complement this comment with the following:
 * This is done by creating a synchronized block on the order object that
 * is passed as parameter to updateState(). The reason the synchronized
 * block is created in the code calling this method is because, usually the
 * order will be inspected before the call to updateState() is made, and
 * this needs to be done within a synchronized block, since several threads
 * may attempt to process the same order at the same time.

src/main/java/org/fogbowcloud/manager/core/datastructures/OrderStateTransitioner.java
```diff
@@ -0,0 +1,73 @@
...
+     *
+     *  @param order, the order that will transition through states
+     * @param newState, the new state
+     */
+    public static boolean updateState(Order order, OrderState newState) {
```
fubica: This method should be called something else. I suggest transitionOrder().

src/main/java/org/fogbowcloud/manager/core/datastructures/OrderStateTransitioner.java
```diff
@@ -0,0 +1,73 @@
...
+        SynchronizedDoublyLinkedList destination = getOrdersList(newState);
+
+        if (origin == null) {
+            LOGGER.error(String.format("Could not find origin list for state %s", currentState));
+        } else if (destination == null) {
+            LOGGER.error(String.format("Could not find destination list for state %s", newState));
+        } else {
+            if (origin.removeItem(order)) {
+                order.setOrderState(newState);
+                origin.addItem(order);
```
fubica: It is not origin, but destination. destination.addItem(order);

src/main/java/org/fogbowcloud/manager/core/datastructures/OrderStateTransitioner.java
fubica: This method should be void. If the order cannot be transitioned it is an indication of a bug. Thus, the method should throw an exception.

src/main/java/org/fogbowcloud/manager/core/datastructures/OrderStateTransitioner.java
```diff
@@ -0,0 +1,73 @@
...
+                String message = String.format(template, order.getId(), currentState.toString());
+                LOGGER.error(message);
+            }
+        }
+
+        return false;
+    }
+
+    protected static SynchronizedDoublyLinkedList getOrdersList(OrderState orderState) {
+        SynchronizedDoublyLinkedList origin = null;
```
fubica: Change the name of the variable origin for list.

src/main/java/org/fogbowcloud/manager/core/datastructures/OrderStateTransitioner.java
arn3tt: Nice! I hadn't see this mistake.

PR #25, opened 2018-05-02
Modify order instance

wassoares:
Create the attribute state in OrderInstance class and removed from the more specific classes


wassoares:
I closed this PR because I created it from my branch with other changes. But now I created a new PR #26.

PR #28, opened 2018-05-07
Exceptions #557

ordansantos:


src/test/java/org/fogbowcloud/manager/core/rest/TestExceptionHandlerController.java
```diff
@@ -0,0 +1,97 @@
...
+		MockHttpServletResponse response = mockMvc.perform(get("/compute/")
+				.accept(MediaType.APPLICATION_JSON))
+        		.andReturn().getResponse();
+		
+		ExceptionResponse r = mapper.readValue(response.getContentAsString(), ExceptionResponse.class);
+		
+		assertEquals(r.getDetails(), "uri=/compute/");
+		assertEquals(r.getMessage(), "Invalid Credentials");
+		assertEquals(r.getStatusCode(), HttpStatus.UNAUTHORIZED);
+		assertEquals(new Integer(response.getStatus()).toString(), HttpStatus.UNAUTHORIZED.toString());
```
llvieira: Is it necessary to instantiate a new `Integer`? I think you can just do `Integer.toString(response.getStatus())`.

src/test/java/org/fogbowcloud/manager/core/plugins/identity/ldap/TestLdapIdentityPlugin.java
```diff
@@ -127,8 +127,8 @@ public void testCreateTokenFail() throws Exception {
 
 			ldapStoneIdentity.createToken(userCredentials);
 			Assert.fail();
-		} catch (Exception e) {
-			Assert.assertEquals(HttpStatus.UNAUTHORIZED.toString(), e.getMessage());
+		} catch (UnauthorizedException e) {
+			Assert.assertEquals("Couldn't load account summary from LDAP Server.", e.getMessage());
```
paulofelipefeitosa: Please set the string constant _"Couldn't load account summary from LDAP Server."_ as a _public static final String_ attribute of the LdapIdentityPlugin class. This change will help to prevent that the test will not crash in case of any modification on these message.

src/main/java/org/fogbowcloud/manager/core/plugins/identity/ldap/LdapIdentityPlugin.java
```diff
@@ -99,8 +103,7 @@ public Token createToken(Map<String, String> userCredentials) {
 			name = ldapAuthenticate(userId, password);
 		} catch (Exception e) {
 			LOGGER.error("Couldn't load account summary from LDAP Server.", e);
-			Integer statusResponse = HttpStatus.SC_UNAUTHORIZED;
-			throw new RuntimeException(statusResponse.toString());
+			throw new InvalidCredentialsException("Couldn't load account summary from LDAP Server.");
```
thiagomanel: sometimes people catch an exception and pass it as an argument to the more specific one. E.g, passing the "e" reference to the InvalidCredentialsException. In this case, there is no need to log. Note that, it is very likely we are going to have two log lines related to this error, one here and another in the code that is going to catch the InvalidCredentials.

I'm not against the current solution but take some time to analyze the trade-offs.

src/main/java/org/fogbowcloud/manager/core/plugins/identity/ldap/LdapIdentityPlugin.java
```diff
@@ -119,10 +122,9 @@ public Token createToken(Map<String, String> userCredentials) {
 			accessId = new String(Base64.encodeBase64(accessId.getBytes(Charsets.UTF_8), false, false), Charsets.UTF_8);
 
 			return new Token(accessId, new Token.User(userId, name), expirationDate, attributes);
-		} catch (Exception e) {
-			LOGGER.error("Erro while trying to sign the token.", e);
-			Integer statusResponse = HttpStatus.SC_UNAUTHORIZED;
-			throw new RuntimeException(statusResponse.toString());
+		} catch (IOException | GeneralSecurityException e) {
```
thiagomanel: related to previous comment

src/main/java/org/fogbowcloud/manager/core/plugins/identity/ldap/LdapIdentityPlugin.java
```diff
@@ -151,15 +153,14 @@ public Token reIssueToken(Token token) {
...
+	public Token getToken(String accessId) throws UnauthorizedException {
 		try {
 			String decodedAccessId = new String(Base64.decodeBase64(accessId), Charsets.UTF_8);
 
 			String split[] = decodedAccessId.split(ACCESSID_SEPARATOR);
 			if (split == null || split.length < 2) {
 				LOGGER.error("Invalid accessID: " + decodedAccessId);
-				Integer statusResponse = HttpStatus.SC_UNAUTHORIZED;
-				throw new RuntimeException(statusResponse.toString());
+				throw new InvalidTokenException("Invalid accessID: " + decodedAccessId);
```
thiagomanel: why both logging and throwing?

src/main/java/org/fogbowcloud/manager/core/plugins/identity/ldap/LdapIdentityPlugin.java
```diff
@@ -170,8 +171,7 @@ public Token getToken(String accessId) {
 
 			if (!verifySign(tokenMessage, signature)) {
 				LOGGER.error("Invalid accessID: " + decodedAccessId);
-				Integer statusResponse = HttpStatus.SC_UNAUTHORIZED;
-				throw new RuntimeException(statusResponse.toString());
+				throw new UnauthorizedException("Invalid accessID: " + decodedAccessId);
```
thiagomanel: why logging and throwing?

src/main/java/org/fogbowcloud/manager/core/rest/ExceptionHandlerController.java
```diff
@@ -0,0 +1,43 @@
...
+
+import org.fogbowcloud.manager.core.exceptions.TokenCreationException;
+import org.fogbowcloud.manager.core.exceptions.UnauthorizedException;
+import org.springframework.http.HttpStatus;
+import org.springframework.http.ResponseEntity;
+import org.springframework.web.bind.annotation.ControllerAdvice;
+import org.springframework.web.bind.annotation.ExceptionHandler;
+import org.springframework.web.context.request.WebRequest;
+import org.springframework.web.servlet.mvc.method.annotation.ResponseEntityExceptionHandler;
+
```
thiagomanel: out of curiosity (I know nothing about rest framework and stuff),  what is the purpose of this class @ordansantos?

src/main/java/org/fogbowcloud/manager/core/rest/ExceptionResponse.java
```diff
@@ -0,0 +1,41 @@
...
+		this.details = details;
+	}
+
+	public HttpStatus getStatusCode() {
+		return statusCode;
+	}
+
+	public void setStatusCode(HttpStatus statusCode) {
+		this.statusCode = statusCode;
+	}
```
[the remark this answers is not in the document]
ordansantos: Good Point. I thought Spring Boot used that for reflection, actually we do need only get methods.

src/main/java/org/fogbowcloud/manager/core/rest/ExceptionHandlerController.java
ordansantos: It maps an exception to an http response.

src/main/java/org/fogbowcloud/manager/core/plugins/identity/ldap/LdapIdentityPlugin.java
ordansantos: I've just fixed it!

src/test/java/org/fogbowcloud/manager/core/plugins/identity/ldap/TestLdapIdentityPlugin.java
ordansantos: We have to discuss more about that, I don't think it's a good idea creating a constant to store this kind of message.

thiagomanel: If you guys do not reach an agreement fast, I suggest to postpone the discussion to the last review (the future PR into develop branch) and accept the code as @ordansantos submitted (obviously, please keep note of @paulofelipefeitosa suggestion somewhere -- maybe in the code itself)

PR #34, opened 2018-05-14
Main initialization

italo-batista:
Implemented code that must runs right after application is up.
- Instantiation and set up of disponibles plugins

### DIFF
--- a/src/main/java/org/fogbowcloud/manager/core/Main.java
+++ b/src/main/java/org/fogbowcloud/manager/core/Main.java
@@ -22,7 +22,7 @@ public class Main implements ApplicationRunner {
 
     private Properties properties;
 
-    private ApplicationController facade;
+    private ApplicationController facade = ApplicationController.getInstance();
     private ManagerController managerController;
     private AuthenticationService authService;
 
@@ -42,7 +42,7 @@ public class Main implements ApplicationRunner {
 
         this.authService = new AuthenticationService(federationIdentityPlugin);
 
-        // TODO: change to use getInstance method since AppCtrl will be Singleton
-        this.facade = new ApplicationController(authService, managerController);
+        this.facade.setAuthenticationController(this.authService);
+        this.facade.setManagerController(this.managerController);
     }
 }
--- a/src/main/java/org/fogbowcloud/manager/core/Main.java
+++ b/src/main/java/org/fogbowcloud/manager/core/Main.java
@@ -1,26 +1,20 @@
 package org.fogbowcloud.manager.core;
 
-import org.fogbowcloud.manager.core.constants.ConfigurationConstants;
 import org.fogbowcloud.manager.core.controllers.ApplicationController;
 import org.fogbowcloud.manager.core.instanceprovider.InstanceProvider;
 import org.fogbowcloud.manager.core.instanceprovider.LocalInstanceProvider;
 import org.fogbowcloud.manager.core.instanceprovider.RemoteInstanceProvider;
 import org.fogbowcloud.manager.core.plugins.IdentityPlugin;
-import org.fogbowcloud.manager.core.plugins.identity.ldap.LdapIdentityPlugin;
+import org.fogbowcloud.manager.core.plugins.compute.ComputePlugin;
 import org.fogbowcloud.manager.core.services.AuthenticationService;
 import org.fogbowcloud.manager.core.services.InstantiationInitService;
-import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.boot.ApplicationArguments;
 import org.springframework.boot.ApplicationRunner;
-import org.springframework.stereotype.Component;
 
 import java.util.Properties;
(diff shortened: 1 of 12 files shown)


src/main/java/org/fogbowcloud/manager/core/plugins/PluginFactory.java
```diff
@@ -0,0 +1,52 @@
...
+import java.lang.reflect.Constructor;
+import java.util.Properties;
+import java.util.logging.Logger;
+
+public class PluginFactory {
+
+    private static final Logger LOGGER = Logger.getLogger(PluginFactory.class.getName());
+    private static final int EXIT_ERROR_CODE = 128;
+
+
```
thiagomanel: @italo-batista why do we need empty constructors?

src/main/java/org/fogbowcloud/manager/core/ManagerController.java
```diff
@@ -41,4 +52,16 @@ private void startManagerThreads() {
 		this.openProcessorThread.start();
 	}
 
-}
\ No newline at end of file
+	public void setComputePlugin(ComputePlugin computePlugin) {
```
thiagomanel: @italo-batista not sure what is the better ... can we pass these object via the constructor instead o setMethods?

src/main/java/org/fogbowcloud/manager/core/services/InstantiationInitService.java
```diff
@@ -0,0 +1,97 @@
...
+import java.io.FileNotFoundException;
+import java.io.IOException;
+import java.util.ArrayList;
+import java.util.List;
+import java.util.Properties;
+import java.util.logging.Logger;
+
+// TODO: search if it is possible locate *.properties in another path
+@Service("instantiationInitService")
+@PropertySource({
```
fgan1: Is it necessary ? The configuration files are been setted in the method seUpProperties.

src/main/java/org/fogbowcloud/manager/core/constants/ConfigurationConstants.java
```diff
@@ -2,11 +2,21 @@
...
+	public static final String MANAGER_CONF_FILE_FULL_PATH = "src/main/resources/manager.properties";
+
+	public static final String INFRA_CONF_FILE_FULL_PATH = "src/main/resources/infrastructure.properties";
+
+	public static final String FEDERATION_CONF_FILE_FULL_PATH = "src/main/resources/federation.properties";
+
+    // XMPP
 	public static final String XMPP_ID_KEY = "xmpp_jid";
 
+	public static final String XMPP_HOST_KEY = "xmpp_host";
```
paulofelipefeitosa: I do not undestand the meaning of this configuration constant.

src/main/java/org/fogbowcloud/manager/core/Main.java
```diff
@@ -0,0 +1,52 @@
...
+
+import java.util.Properties;
+
+public class Main implements ApplicationRunner {
+
+    private InstantiationInitService instantiationInitService;
+
+    private Properties properties;
+
+    private ApplicationController facade;
```
GustavoYamaguchi: This variable is not being used

src/main/java/org/fogbowcloud/manager/core/Main.java
```diff
@@ -0,0 +1,52 @@
...
+import org.fogbowcloud.manager.core.plugins.IdentityPlugin;
+import org.fogbowcloud.manager.core.plugins.compute.ComputePlugin;
+import org.fogbowcloud.manager.core.services.AuthenticationService;
+import org.fogbowcloud.manager.core.services.InstantiationInitService;
+import org.springframework.boot.ApplicationArguments;
+import org.springframework.boot.ApplicationRunner;
+
+import java.util.Properties;
+
+public class Main implements ApplicationRunner {
```
GustavoYamaguchi: This class seems to not work. Please, check this out.

src/main/java/org/fogbowcloud/manager/core/constants/ConfigurationConstants.java
italo-batista: Is it the XMPP server IP address. I was using in an old test, but no need using anymore.

src/main/java/org/fogbowcloud/manager/core/plugins/PluginFactory.java
italo-batista: No need here. I got too _addicted_ to empty constructors. Sorry.

src/main/java/org/fogbowcloud/manager/core/ManagerController.java
italo-batista: I wanted to avoid that the ManagerController's constructor was too big, since there will be a lot of plugins. But I recognize your suggestion is safer. I will change it.

PR #36, opened 2018-05-14
Create ClosedOrdersProcessor and test it

arn3tt:


src/main/java/org/fogbowcloud/manager/core/threads/ClosedComputeOrdersProcessor.java
```diff
@@ -0,0 +1,110 @@
...
+                }
+            } else {
+                LOGGER.info(String.format("Order %s has no instance associated", order.getId()));
+            }
+        }
+    }
+
+    private void deactivateOrder(Order order) {
+        Map<String, Order> activeOrdersMap = sharedOrderHolders.getActiveOrdersMap();
+        if (activeOrdersMap.get(order.getId()) != null) {
```
paulofelipefeitosa: Is the containsKey method from Map more apropriated?

src/test/java/org/fogbowcloud/manager/core/threads/TestClosedComputeOrdersProcessor.java
```diff
@@ -0,0 +1,491 @@
...
+    public void testProcessClosedRemoteOrder() {
+
+    }
+
+    @Test
+    public void testProcessClosedOrderFails() {
+
+    }
+
+//    /**
```
paulofelipefeitosa: Why theses tests are commented?

src/main/java/org/fogbowcloud/manager/core/threads/ClosedComputeOrdersProcessor.java
```diff
@@ -0,0 +1,110 @@
...
+    @Override
+    public void run() {
+        boolean isActive = true;
+        while (isActive) {
+            try {
+                Order order = this.closedOrders.getNext();
+                if (order != null) {
+                    try {
+                        processClosedOrder(order);
+                        closedOrders.removeItem(order);
```
paulofelipefeitosa: Why you are removing the order from closed list in the run method?

src/test/java/org/fogbowcloud/manager/core/threads/TestClosedComputeOrdersProcessor.java
arn3tt: I forgot to remove them, will fix this.

src/main/java/org/fogbowcloud/manager/core/threads/ClosedComputeOrdersProcessor.java
arn3tt: Please take a look at the last page of our Design Contract. The pseudo-code of this member is detailed there.

src/main/java/org/fogbowcloud/manager/core/threads/ClosedComputeOrdersProcessor.java
arn3tt: Nice catch! Will fix this in the next commits.

arn3tt:
@paulofelipefeitosa can you review this PR again? I fixed and discussed what you found.

src/main/java/org/fogbowcloud/manager/core/threads/ClosedComputeOrdersProcessor.java
paulofelipefeitosa: Sorry, you do not understand my question, I know that is necessary to remove the order from the closed orders list, I am questioning why you are doing it at the run method? Is more appropriated to do it at the process method?

PR #39, opened 2018-05-16
Delete compute #545

wassoares:
DONE:
- Implementation of task DELETE compute # 545 with tests of delete method in 'ApplicationController' and 'ComputeOrdersService' classes;


paulofelipefeitosa:
@arn3tt, please, review this PR.

src/main/java/org/fogbowcloud/manager/core/rest/controllers/ComputeOrdersController.java
```diff
@@ -57,7 +59,9 @@
 	}
 
 	@RequestMapping(value = "/{id}", method = RequestMethod.DELETE)
-	public ResponseEntity<Boolean> deleteCompute(@PathVariable String id) {
+	public ResponseEntity<Boolean> deleteCompute(@PathVariable String id,
+			@RequestHeader(value = "accessId") String accessId) throws UnauthorizedException, OrderManagementException {
```
llvieira: Rather than use `"accessId"` you can use the constant `ACCESS_ID_HEADER_KEY`. Why `@delete` does not have the `localTokenId` as a header?

src/main/java/org/fogbowcloud/manager/core/rest/controllers/ComputeOrdersController.java
```diff
@@ -57,7 +59,9 @@
 	}
 
 	@RequestMapping(value = "/{id}", method = RequestMethod.DELETE)
-	public ResponseEntity<Boolean> deleteCompute(@PathVariable String id) {
+	public ResponseEntity<Boolean> deleteCompute(@PathVariable String id,
```
llvieira: You can add some log message at this method.

PR #43, opened 2018-05-16
Local Instance Provider

arn3tt:


src/main/java/org/fogbowcloud/manager/core/instanceprovider/InstanceProvider.java
```diff
@@ -1,13 +1,14 @@
...
+import org.fogbowcloud.manager.core.models.orders.ComputeOrder;
 import org.fogbowcloud.manager.core.models.orders.Order;
 import org.fogbowcloud.manager.core.models.orders.instances.OrderInstance;
 import org.fogbowcloud.manager.core.models.token.Token;
 
 public interface InstanceProvider {
 
 	/**
-	 * Request an Instance for an Order.
+	 * Requests an Instance for a ComputeOrder.
```
paulofelipefeitosa: This is a general implementation.

src/main/java/org/fogbowcloud/manager/core/instanceprovider/LocalInstanceProvider.java
```diff
@@ -0,0 +1,42 @@
...
+            throw new UnsupportedOperationException("Not implemented yet.");
+        } else {
+            throw new UnsupportedOperationException("Not implemented yet.");
+        }
+    }
+
+    private OrderInstance requestInstance(ComputeOrder order) throws RequestException {
+        String imageName = order.getImageName();
+        String id = computePlugin.requestInstance(order, imageName);
+        return new OrderInstance(id, InstanceState.ACTIVE);
```
paulofelipefeitosa: The InstanceState should be inactive.

arn3tt: Nice catch, I will fix this soon.

src/main/java/org/fogbowcloud/manager/core/instanceprovider/InstanceProvider.java
arn3tt: Right, will fix this.

PR #71, opened 2018-06-22
Resolving integration problems

ordansantos:


src/main/java/org/fogbowcloud/manager/Main.java
```diff
@@ -52,6 +63,13 @@ public void run(ApplicationArguments args) {
 
         XmppComponentManager xmppComponentManager = new XmppComponentManager(xmppJid, xmppPassword, xmppServerIp,
                 xmppServerPort, xmppTimeout);
+        
+        try {
+			xmppComponentManager.connect();
+		} catch (ComponentException e) {
+			LOGGER.error("Unable to connect to XMPP, check XMPP configuration file");
+			return;
```
paulofelipefeitosa: Do you think that is better to use System.exit with an error status code? Using return the program will exit with status code 0 (success).

PR #72, opened 2018-06-26
Networks id list

wassoares:
- Added attribute list 'networksId' in  'ComputeOrder' class, making modifications to 'OpenStackNovaV2ComputePlugin' class, and fixed tests;

### DIFF
--- a/src/main/java/org/fogbowcloud/manager/core/models/orders/ComputeOrder.java
+++ b/src/main/java/org/fogbowcloud/manager/core/models/orders/ComputeOrder.java
@@ -1,5 +1,6 @@
 package org.fogbowcloud.manager.core.models.orders;
 
+import java.util.Collections;
 import java.util.List;
 import java.util.UUID;
 
@@ -42,7 +43,8 @@ public class ComputeOrder extends Order {
             int disk,
             String imageId,
             UserData userData,
-            String publicKey, List<String> networksId) {
+            String publicKey, 
+            List<String> networksId) {
         super(id, federationUser, requestingMember, providingMember);
         this.vCPU = vCPU;
         this.memory = memory;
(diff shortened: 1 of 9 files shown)


src/main/java/org/fogbowcloud/manager/core/models/orders/ComputeOrder.java
```diff
@@ -39,14 +42,15 @@ public ComputeOrder(
             int disk,
             String imageId,
             UserData userData,
-            String publicKey) {
+            String publicKey, List<String> networksId) {
```
GustavoYamaguchi: Since the parameters have a line break after each one, I think you should also add a line break after publickKey parameter

src/main/java/org/fogbowcloud/manager/core/plugins/cloud/openstack/OpenStackNovaV2ComputePlugin.java
```diff
@@ -163,8 +163,16 @@ protected String getTenantId(Token localToken) {
         return tokenAttr.get(TENANT_ID);
     }
 
-    protected String getNetworkId() {
-        return this.properties.getProperty(DEFAULT_NETWORK_ID_KEY);
+    protected List<String> getNetworksId(ComputeOrder computeOrder) {
+        List<String> networksId = computeOrder.getNetworksId();
+        String id = this.properties.getProperty(DEFAULT_NETWORK_ID_KEY);
```
GustavoYamaguchi: I think 'id' is not a good name for this variable. I'd use defaultNetworkId

src/main/java/org/fogbowcloud/manager/core/models/orders/ComputeOrder.java
```diff
@@ -115,4 +121,8 @@ public String getPublicKey() {
         return publicKey;
     }
 
+    public List<String> getNetworksId() {
+        return networksId;
```
GustavoYamaguchi: I think that would be better to return a Collections.unmodifiableList(networksId), and have a set method for networksId, instead of returning the list itself.

PR #75, opened 2018-07-05
History & Recovery

llvieira:


src/main/java/org/fogbowcloud/manager/core/datastore/orderstorage/OrderTimestampStorage.java
```diff
@@ -0,0 +1,65 @@
+package org.fogbowcloud.manager.core.datastore.orderstorage;
+
+import org.apache.log4j.Logger;
+import org.fogbowcloud.manager.core.datastore.commands.TimestampSQLCommands;
+import org.fogbowcloud.manager.core.models.orders.Order;
+
+import java.sql.*;
+import java.util.Date;
+
```
thiagomanel: add a comment here why do we need this special *TimestampStorage class. What is the feature we added?

src/main/java/org/fogbowcloud/manager/core/models/orders/OrderState.java
```diff
@@ -1,10 +1,33 @@
...
+
+    OrderState(String value) {
+        this.value = value;
+    }
+
+    @JsonValue
+    public String getValue() {
+        return value;
+    }
+
```
thiagomanel: doesn't enum already has something like this fromValue method?

src/main/java/org/fogbowcloud/manager/core/datastore/DatabaseManager.java
```diff
@@ -0,0 +1,99 @@
...
+        this.orderTimestampStorage = new OrderTimestampStorage();
+    }
+
+    public static DatabaseManager getInstance() {
+        if (instance == null) {
+            try {
+                instance = new DatabaseManager();
+            } catch (SQLException e) {
+                LOGGER.error("Error instantiating database manager", e);
+                throw new Error(e);
```
thiagomanel: We now have a recipe about how to throw exception. We can throw a specified subset. Are you following this? In other words, are we allowed to throw this "new Error" exception?

src/main/java/org/fogbowcloud/manager/core/datastore/commands/OrderTableAttributes.java
```diff
@@ -0,0 +1,65 @@
+package org.fogbowcloud.manager.core.datastore.commands;
+
+public class OrderTableAttributes {
+
```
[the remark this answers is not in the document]
llvieira: I think this database do not know anything about fednet.

src/main/java/org/fogbowcloud/manager/core/models/orders/OrderState.java
llvieira: It doesn't.

src/main/java/org/fogbowcloud/manager/core/datastore/DatabaseManager.java
llvieira: Using `FatalErrorException` now, since it is a runtime exception.

src/main/java/org/fogbowcloud/manager/core/datastore/commands/OrderTableAttributes.java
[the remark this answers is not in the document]
thiagomanel: ok

src/main/java/org/fogbowcloud/manager/core/models/orders/OrderState.java
thiagomanel: how about the OrderState.valueOf? Doesn't this do that?

src/main/java/org/fogbowcloud/manager/core/datastore/orderstorage/OrderTimestampStorage.java
llvieira: Done!

src/main/java/org/fogbowcloud/manager/core/models/orders/OrderState.java
llvieira: Fixed! I can't use this before because the Enum had no value. Thanks!

src/main/java/org/fogbowcloud/manager/core/OrderStateTransitioner.java
```diff
@@ -85,7 +90,13 @@ public static void deactivateOrder(Order order) throws UnexpectedException {
                         "Tried to remove order %s from the active orders but it was not active", order.getId());
                 throw new UnexpectedException(message);
             }
+
             closedOrders.removeItem(order);
+
+            // TODO: is it here? (only used for closed orders, here the resource is definitely deleted on provider)
+            order.setInstanceId(null);
```
thiagomanel: @llvieira I did't quite understand why not adding the setInstance(null) in the CloseProcessor instead of here (there is at least one advantage of doing there: we release cloud resources in the CloseProcessor)

src/main/java/org/fogbowcloud/manager/core/OrderStateTransitioner.java
```diff
@@ -110,6 +121,14 @@ private static void doTransition(Order order, OrderState newState) throws Unexpe
         } else {
             if (origin.removeItem(order)) {
                 order.setOrderState(newState);
+
+                // TODO: Updating order in stable storage
+                // (do not update closed orders here since it still have an instance id)
```
thiagomanel: I did not understand why not updating closed order in the database (I see no one cares about closed resources, but if we do not introduce a bug in case we do update it, it seems better to avoid one more if in the code)

src/main/java/org/fogbowcloud/manager/core/SharedOrderHolders.java
```diff
@@ -20,13 +23,30 @@
...
+        this.spawningOrders = databaseManager.readActiveOrders(OrderState.SPAWNING);
+        addOrdersToMap(this.spawningOrders, this.activeOrdersMap);
+        this.failedOrders = databaseManager.readActiveOrders(OrderState.FAILED);
+        addOrdersToMap(this.failedOrders, this.activeOrdersMap);
+        this.fulfilledOrders = databaseManager.readActiveOrders(OrderState.FULFILLED);
+        addOrdersToMap(this.fulfilledOrders, this.activeOrdersMap);
+        this.pendingOrders = databaseManager.readActiveOrders(OrderState.PENDING);
+        addOrdersToMap(this.pendingOrders, this.activeOrdersMap);
+        this.closedOrders = databaseManager.readActiveOrders(OrderState.CLOSED);
+        addOrdersToMap(this.closedOrders, this.activeOrdersMap);
```
thiagomanel: what is the idiom of using this List? Are one supposed to reset it whenever it knows the end was reached?

src/main/java/org/fogbowcloud/manager/core/datastore/DatabaseManager.java
```diff
@@ -0,0 +1,100 @@
...
+        this.orderTimestampStorage = new OrderTimestampStorage();
+    }
+
+    public static DatabaseManager getInstance() {
+        if (instance == null) {
+            try {
+                instance = new DatabaseManager();
+            } catch (SQLException e) {
+                LOGGER.error("Error instantiating database manager", e);
+                throw new FatalErrorException(e.getMessage());
```
thiagomanel: pass the exception as param

src/main/java/org/fogbowcloud/manager/core/datastore/DatabaseManager.java
```diff
@@ -0,0 +1,100 @@
...
+                break;
+            case ATTACHMENT:
+                this.attachmentOrderStorage.addOrder(order);
+                this.orderTimestampStorage.addOrder(order);
+                break;
+        }
+    }
+
+    @Override
+    public void update(Order order) {
```
thiagomanel: what if this order was not added to the database?

src/main/java/org/fogbowcloud/manager/core/datastore/orderstorage/AttachmentOrderStorage.java
```diff
@@ -0,0 +1,163 @@
...
+            connection.commit();
+        } catch (SQLException e) {
+            LOGGER.error("Couldn't add the attachment order.", e);
+            try {
+                if (connection != null) {
+                    connection.rollback();
+                }
+            } catch (SQLException e1) {
+                LOGGER.error("Couldn't rollback transaction.", e1);
+            }
```
thiagomanel: you missed the thrown

src/main/java/org/fogbowcloud/manager/core/datastore/orderstorage/OrderStorage.java
```diff
@@ -0,0 +1,87 @@
...
+    }
+
+    protected void closeConnection(Statement statement, Connection connection) {
+        if (statement != null) {
+            try {
+                if (!statement.isClosed()) {
+                    statement.close();
+                }
+            } catch (SQLException e) {
+                LOGGER.error("Couldn't close statement");
```
thiagomanel: add the exception to the log call

llvieira: Fixed!

src/main/java/org/fogbowcloud/manager/core/OrderStateTransitioner.java
llvieira: The method `deactivateOrder(Order order)` where that piece of code is located is just called by the `ClosedProcessor`. Therefore, it already works fine!

src/main/java/org/fogbowcloud/manager/core/OrderStateTransitioner.java
llvieira: I was not updating closed orders here since it already has been updated in `deactivateOrder(Order order)` method, so we could not add an order with same `ID` and `OrderState` twice in the `timestamp` table. However, now we created the`DEACTIVATE` status in which solve this problem.

So, an order that has been closed is stored twice in stable storage: one when the order is deleted (but instanceId is still not null), and another when it is deactivated (when instanceId is null); we need the deactivate state so that the add in the timestamp table won't break.

src/main/java/org/fogbowcloud/manager/core/SharedOrderHolders.java
llvieira: I don't got it.

src/main/java/org/fogbowcloud/manager/core/datastore/DatabaseManager.java
llvieira: I'm not sure about that, yet.

src/main/java/org/fogbowcloud/manager/core/SharedOrderHolders.java
fubica: What Thiago means is that once you have iterated in the list until its end, you should reset the pointer, so that the next thread that iterates in the list will start from the beginning. This is just a performance improvement.

src/main/java/org/fogbowcloud/manager/core/datastore/DatabaseManager.java
fubica: Normally, this should never happen, since update() is only called to an order that has already been activated, and the order is added to the db when it is activated. As I understood, whenever an operation in the db fails, we simply log an error. We will need to develop a tool that reads the log and notifies the site admin of these and other kinds of problems.

src/main/java/org/fogbowcloud/manager/core/SharedOrderHolders.java
llvieira: Done!

src/main/java/org/fogbowcloud/manager/core/datastore/DatabaseManager.java
llvieira: Yes, currently we are not raising exceptions when operations in the db are performed. We must decide whether logging it's enough.

src/main/java/org/fogbowcloud/manager/core/datastore/orderstorage/AttachmentOrderStorage.java
llvieira: See: https://github.com/fogbow/fogbow-manager-core/pull/75#discussion_r201310553

src/main/java/org/fogbowcloud/manager/core/SharedOrderHolders.java
thiagomanel: ok

src/main/java/org/fogbowcloud/manager/core/datastore/DatabaseManager.java
thiagomanel: A general comment: I follow fubica's advice of logging the database errors. However, I'm a little bit afraid of keep system operating without the database (I'm trying to devise the sysadmin procedure in case of such failures; as the pain may be high, sometimes is better to limit the damage).

In summary, @llvieira keep the code as it is now (logging the error instead of killing the core). also, keep in mind the admin tool is important, so we could try to develop it as soon as possible

fubica: Note that the problems with running without stable storage are twofold. Firstly, if there is a crash, I won't be able to recover the system state. This should be rare, and shouldn't be a big issue. Secondly, I won't have accounting information registered. This is potentially more harmful, but should we stop the system to avoid this? I don't think so. I think that the best we can do is to equip the system administrator with tools that will allow her/him to identify these problems as soon as possible, and will fix them accordingly. There are also problems that may happen even if the database is operational. Something may change in the system and a crash may occur before these changes are stored in stable storage. Again, the system administrator needs to come in play to fix possible inconsistencies. I agree that this tool should be develop asap.

src/main/java/org/fogbowcloud/manager/core/constants/ConfigurationConstants.java
```diff
@@ -40,4 +40,7 @@
     public static final String REVERSE_TUNNEL_PRIVATE_ADDRESS_KEY = "reverse_tunnel_private_address";
     public static final String REVERSE_TUNNEL_PORT_KEY = "reverse_tunnel_port";
     public static final String REVERSE_TUNNEL_HTTP_PORT_KEY = "reverse_tunnel_http_port";
+
+    // HISTORY & RECOVERY DATABASE CONF
+    public static final String DATABASE_URL = "history_recovery_database_url";
```
paulofelipefeitosa: Thinking as a fogbow user, I do not think that this property name is representative.

llvieira: Okay, I will change it to `manager_jdbc_url` as we discussed.

paulofelipefeitosa: Thank you Viga :top:! This name solves the misunderstand problem with others databases urls :smile:.

llvieira: Exactly, thank you too, hehe :man_technologist:.

src/main/java/org/fogbowcloud/manager/core/datastore/DatabaseManager.java
llvieira: Agreed, @thiagomanel and @fubica. To stop the whole system in consequence of some database failures really seems a bad idea. In regard to the new tool, I'm willing to help with it.

llvieira: Agreed, @thiagomanel and @fubica. Stopping the whole system in consequence of some database failures really seems a bad idea. In regard to the new tool, I'm willing to help with it.

src/main/java/org/fogbowcloud/manager/core/constants/ConfigurationConstants.java
fubica: Call it jdbc_manager_url instead.

paulofelipefeitosa: @fubica `jdbc_manager_url` can be understood as the manager url in jdbc.

fubica: I see the point now. At first, I didn't understand that manager was a reference to fogbow manager. I wonder if we need to have manager in the variable name. Do we have other jdbc_url in this config file? Perhaps we could just call it jdbc_url.

paulofelipefeitosa: In the manager I don't think so, but in others services with databases, such as the federated network service, certainly there will be a `jdbc_url` constant, but these constants will be separated by conf files.

llvieira: I think the name could be just `jdbc_database_url`, since it represents the database url defined by the jdbc patterns.

fubica:
Agreed.

Em Qua, 11 de jul de 2018 14:12, Lucas Lima Vieira <notifications@github.com>
escreveu:

PR #79, opened 2018-07-12
Coverage local cloud connector

jvmafra:


src/test/java/org/fogbowcloud/manager/core/cloudconnector/LocalCloudConnectorTest.java
```diff
@@ -0,0 +1,359 @@
...
+	    
+	    assertEquals(1, quota.getUsedQuota().getvCPU());
+	    assertEquals(1024, quota.getUsedQuota().getRam());
+	    assertEquals(1, quota.getUsedQuota().getInstances());
+	}
+	
+	// If the instance type isn't of Compute type, an exception must be throw
+	@Test(expected = UnexpectedException.class)
+	public void testGetUserQuotaException() throws FogbowManagerException, UnexpectedException {    
+	    this.localCloudConnector.getUserQuota(federationUser, InstanceType.VOLUME);
```
thiagomanel: @jvmafra maybe add the other types

src/test/java/org/fogbowcloud/manager/core/cloudconnector/LocalCloudConnectorTest.java
```diff
@@ -0,0 +1,359 @@
...
+	@Test
+	public void testGetUserComputeQuota() throws FogbowManagerException, UnexpectedException {
+		
+		ComputeAllocation fakeTotalComputeAllocation = new ComputeAllocation(2, 2048, 2);
+		ComputeAllocation fakeUsedComputeAllocation = new ComputeAllocation(1, 1024, 1);
+		ComputeQuota fakeComputeQuota = new ComputeQuota(fakeTotalComputeAllocation, fakeUsedComputeAllocation);
+	    when(this.computeQuotaPlugin.getUserQuota(any(Token.class))).thenReturn(fakeComputeQuota);
+	    
+	    ComputeQuota quota = (ComputeQuota) this.localCloudConnector.getUserQuota(federationUser, InstanceType.COMPUTE);
+	    assertEquals(2, quota.getTotalQuota().getvCPU());
```
thiagomanel: instead of using literal, e.g 2, 2018 and 2 in line 318 and 325, replace them by a variable with a good name

src/test/java/org/fogbowcloud/manager/core/cloudconnector/LocalCloudConnectorTest.java
```diff
@@ -0,0 +1,359 @@
...
+	@Test
+	public void testDeleteAttachmentInstance() throws FogbowManagerException, UnexpectedException {
+		this.order = Mockito.mock(AttachmentOrder.class);
+	    when(this.order.getType()).thenReturn(InstanceType.ATTACHMENT);
+	    when(this.order.getInstanceId()).thenReturn(FAKE_INSTANCE_ID);
+		this.localCloudConnector.deleteInstance(order);
+	}
+	
+	@Test
+	public void testgetImage() throws FogbowManagerException, UnexpectedException {
```
thiagomanel: follow the camel case naming standard

src/test/java/org/fogbowcloud/manager/core/cloudconnector/LocalCloudConnectorTest.java
```diff
@@ -0,0 +1,359 @@
...
+	    when(this.order.getInstanceId()).thenReturn(FAKE_INSTANCE_ID);
+		this.localCloudConnector.deleteInstance(order);
+	}
+	
+	@Test
+	public void testDeleteAttachmentInstance() throws FogbowManagerException, UnexpectedException {
+		this.order = Mockito.mock(AttachmentOrder.class);
+	    when(this.order.getType()).thenReturn(InstanceType.ATTACHMENT);
+	    when(this.order.getInstanceId()).thenReturn(FAKE_INSTANCE_ID);
+		this.localCloudConnector.deleteInstance(order);
```
thiagomanel: I did not understand the assert condition in this test. where is it?

src/test/java/org/fogbowcloud/manager/core/cloudconnector/LocalCloudConnectorTest.java
```diff
@@ -0,0 +1,359 @@
...
+	}
+	
+	@Test
+	public void testDeleteNetworkInstance() throws FogbowManagerException, UnexpectedException {
+		this.order = Mockito.mock(NetworkOrder.class);
+	    when(this.order.getType()).thenReturn(InstanceType.NETWORK);
+	    when(this.order.getInstanceId()).thenReturn(FAKE_INSTANCE_ID);
+		this.localCloudConnector.deleteInstance(order);
```
thiagomanel: same comment as in testDeleteAttachmentInstance

src/test/java/org/fogbowcloud/manager/core/cloudconnector/LocalCloudConnectorTest.java
```diff
@@ -0,0 +1,359 @@
...
+	}
+	
+	@Test
+	public void testDeleteVolumeInstance() throws FogbowManagerException, UnexpectedException {
+		this.order = Mockito.mock(VolumeOrder.class);
+	    when(this.order.getType()).thenReturn(InstanceType.VOLUME);
+	    when(this.order.getInstanceId()).thenReturn(FAKE_INSTANCE_ID);
+		this.localCloudConnector.deleteInstance(order);
```
thiagomanel: same comment as in testDeleteAttachmentInstance

src/test/java/org/fogbowcloud/manager/core/cloudconnector/LocalCloudConnectorTest.java
```diff
@@ -0,0 +1,359 @@
...
+	    //when(computePlugin.requestInstance(any(ComputeOrder.class), any(Token.class))).thenReturn(FAKE_INSTANCE_ID);
+		this.localCloudConnector.deleteInstance(order);
+	}
+	
+	@Test
+	public void testDeleteComputeInstance() throws FogbowManagerException, UnexpectedException {
+		this.order = Mockito.mock(ComputeOrder.class);
+	    when(this.order.getType()).thenReturn(InstanceType.COMPUTE);
+	    when(this.order.getInstanceId()).thenReturn(FAKE_INSTANCE_ID);
+		this.localCloudConnector.deleteInstance(order);
```
thiagomanel: same comment from testDeleteAttachmentInstance

src/test/java/org/fogbowcloud/manager/core/cloudconnector/LocalCloudConnectorTest.java
```diff
@@ -0,0 +1,359 @@
...
+	    when(this.order.getId()).thenReturn(FAKE_ORDER_ID);
+		assertEquals(FAKE_ORDER_ID, this.localCloudConnector.getInstance(order).getId());
+	}
+	
+	@Test
+	public void testDeleteInstanceWithoutInstanceID() throws FogbowManagerException, UnexpectedException {
+		this.order = Mockito.mock(ComputeOrder.class);
+	    when(this.order.getType()).thenReturn(InstanceType.COMPUTE);
+	    //when(computePlugin.requestInstance(any(ComputeOrder.class), any(Token.class))).thenReturn(FAKE_INSTANCE_ID);
+		this.localCloudConnector.deleteInstance(order);
```
thiagomanel: what is the assert condition? 
delete this commented line (267)

src/test/java/org/fogbowcloud/manager/core/cloudconnector/LocalCloudConnectorTest.java
```diff
@@ -0,0 +1,359 @@
...
+	@Test
+	public void testgetEmptyAttachmentInstance() throws FogbowManagerException, UnexpectedException {
+		this.order = Mockito.mock(AttachmentOrder.class);
+	    when(this.order.getType()).thenReturn(InstanceType.ATTACHMENT);
+	    when(this.order.getInstanceId()).thenReturn(null);
+	    when(this.order.getId()).thenReturn(FAKE_ORDER_ID);
+		assertEquals(FAKE_ORDER_ID, this.localCloudConnector.getInstance(order).getId());
+	}
+	
+	// The order doesn't have an InstanceID, so the method an empty ComputeInstance is returned with the same id of order.
```
thiagomanel: it is very pleasant to have a small description (line 253) of test behaviour. sometimes, mocked stuff is hard to read.

src/test/java/org/fogbowcloud/manager/core/cloudconnector/LocalCloudConnectorTest.java
```diff
@@ -0,0 +1,359 @@
...
+	@Test
+	public void testgetEmptyComputeInstance() throws FogbowManagerException, UnexpectedException {
```
thiagomanel: follow camel case naming @jvmafra

src/test/java/org/fogbowcloud/manager/core/cloudconnector/LocalCloudConnectorTest.java
```diff
@@ -0,0 +1,359 @@
...
+	// If plugin returns a null instance id, the method requestInstance() must throw an exception
+	@Test(expected = UnexpectedException.class)
+	public void testExceptionNullInstanceId() throws FogbowManagerException, UnexpectedException {
+		this.order = Mockito.mock(ComputeOrder.class);
+	    when(this.order.getType()).thenReturn(InstanceType.COMPUTE);
+	    when(computePlugin.requestInstance(any(ComputeOrder.class), any(Token.class))).thenReturn(null);
+		this.localCloudConnector.requestInstance(order);
+	}
+	
+
```
thiagomanel: follow camel case ...

src/test/java/org/fogbowcloud/manager/core/cloudconnector/LocalCloudConnectorTest.java
jvmafra: I didn't get it. There is a problem in description? 
I've fixed the description by eliminating "the method": "The order doesn't have an InstanceID, so an empty ComputeInstance is returned with the same id of order."

src/test/java/org/fogbowcloud/manager/core/cloudconnector/LocalCloudConnectorTest.java
jvmafra: Done

src/test/java/org/fogbowcloud/manager/core/cloudconnector/LocalCloudConnectorTest.java
jvmafra: Done

src/test/java/org/fogbowcloud/manager/core/cloudconnector/LocalCloudConnectorTest.java
jvmafra: Done

src/test/java/org/fogbowcloud/manager/core/cloudconnector/LocalCloudConnectorTest.java
jvmafra: deleteInstance() method does nothing (just log). It delegates the task to the plugin. I call the method to make sure that any error happens. I can delete the test (decreasing the coverage)  or I can use a try/catch block with Assert.fail() into catch.

src/test/java/org/fogbowcloud/manager/core/cloudconnector/LocalCloudConnectorTest.java
thiagomanel: no @jvmafra I said it nice to add such description. keep adding them

src/test/java/org/fogbowcloud/manager/core/cloudconnector/LocalCloudConnectorTest.java
thiagomanel: @jvmafra I think you need to modify this method and also the testDeleteInstanceWithoutInstanceID, testDeleteComputeInstance, testDeleteVolumeInstance, testDeleteNetworkInstance, testDeleteAttachmentInstance to assert something. In this case, the assert is the forward of the call to the plugin. this is the expected behaviour. After that, I think the PR is ok

src/test/java/org/fogbowcloud/manager/core/cloudconnector/LocalCloudConnectorTest.java
thiagomanel: also, please, test this nullity-exception with the order plugins, not only with the compute one

src/test/java/org/fogbowcloud/manager/core/cloudconnector/LocalCloudConnectorTest.java
```diff
@@ -0,0 +1,382 @@
...
+		this.order = Mockito.mock(ComputeOrder.class);
+	    when(this.order.getType()).thenReturn(InstanceType.COMPUTE);
+	    when(this.order.getInstanceId()).thenReturn(FAKE_INSTANCE_ID);
+	    when(computePlugin.getInstance(any(String.class), any(Token.class))).thenReturn(this.computeInstance);
+		assertEquals(FAKE_INSTANCE_ID, localCloudConnectorSpy.getInstance(order).getId());
+	}
+	
+	
+	// The order doesn't have an InstanceID, so an empty NetworkInstance is returned with the same id of order.
```
thiagomanel: I'm no sure, maybe we need to check this Instances are "empty" @jvmafra talk to @fgan1 to discuss about it (I'm inclined to say it is ok, already)

jvmafra: I had doubts in this method. The description says "When there is no instance, an empty one is created with the appropriate state", but the new instance is created only with id, without state. We can discuss on monday if this behavior is correct. 

About the test, we really can check if the instance is empty. All de attributes must be null, except id.

src/test/java/org/fogbowcloud/manager/core/cloudconnector/LocalCloudConnectorTest.java
jvmafra: Done

src/test/java/org/fogbowcloud/manager/core/cloudconnector/LocalCloudConnectorTest.java
jvmafra: Done

src/test/java/org/fogbowcloud/manager/core/cloudconnector/LocalCloudConnectorTest.java
```diff
@@ -0,0 +1,557 @@
...
+	public void testExceptionNullVolumeInstanceId() throws FogbowManagerException, UnexpectedException {
+		this.order = Mockito.mock(VolumeOrder.class);
+	    when(this.order.getType()).thenReturn(InstanceType.VOLUME);
+	    when(volumePlugin.requestInstance(any(VolumeOrder.class), any(Token.class))).thenReturn(null);
+		this.localCloudConnector.requestInstance(order);
+	}
+	
+	// The order has an InstanceID, so the method getResourceInstance() is called.
+	@Test
+	public void testGetNetworkInstance() throws FogbowManagerException, UnexpectedException {
```
thiagomanel: where is the expected getResourceInstance behaviour configured, @jvmafra ?

jvmafra: when(networkPlugin.getInstance(any(String.class), any(Token.class))).thenReturn(this.networkInstance);

getResourceInstance calls the method getInstance of plugin and we are mocking it.

jvmafra: The documentation tries to say that the if branch will be executed because the order has an instance id

PR #80, opened 2018-07-12
Created tests for OpenstackStateMapper 

llvieira:


src/test/java/org/fogbowcloud/manager/core/plugins/cloud/openstack/OpenStackStateMapperTest.java
```diff
@@ -0,0 +1,317 @@
+package org.fogbowcloud.manager.core.plugins.cloud.openstack;
+
+import org.fogbowcloud.manager.core.models.instances.InstanceState;
+import org.fogbowcloud.manager.core.models.instances.InstanceType;
+import org.junit.Assert;
+import org.junit.Test;
+
+public class OpenStackStateMapperTest {
+
```
thiagomanel: @llvieira one general comment (related to the code not to the tests): i) do you know how coupled with an specific openstack API are we? (I mean, we are parsing string from openstack API, right? does this chance often? are they the same on the past N releases?)

thiagomanel: I saw compute default lead to InstanceState.FAILED, network default to InstanceState.INCONSISTENT, volume and attachment do not have a default. Is it correct? (i'm afraid of the lack of symmetry ... I have not against the code per se)

src/test/java/org/fogbowcloud/manager/core/plugins/cloud/openstack/OpenStackStateMapperTest.java
```diff
@@ -0,0 +1,317 @@
...
+    public void testComputeActiveStatusToReady() {
+        InstanceType instanceType = InstanceType.COMPUTE;
+
+        InstanceState instanceState1 = OpenStackStateMapper.map(instanceType, "active");
+        InstanceState instanceState2 = OpenStackStateMapper.map(instanceType, "ACTIVE");
+
+        Assert.assertEquals(instanceState1, InstanceState.READY);
+        Assert.assertEquals(instanceState2, InstanceState.READY);
+    }
+
```
thiagomanel: @llvieira should we protect the code agains a null parameter? the code is throwing a null point exception in such case.

src/test/java/org/fogbowcloud/manager/core/plugins/cloud/openstack/OpenStackStateMapperTest.java
llvieira: @thiagomanel what I realize was that OpenStack keeps the same status through different versions, although they can add new ones. Also, in regard to our scenario I just found out different OpenStack API's versions to volume. 

Volume status table in two OpenStack API versions:
https://developer.openstack.org/api-ref/block-storage/v2/#volumes-volumes
https://developer.openstack.org/api-ref/block-storage/v3/#volumes-volumes **[current]**

Compute status information:
https://developer.openstack.org/api-ref/compute/?#list-servers

Network status information:
https://developer.openstack.org/api-ref/network/v2/#show-network-details

I think @fubica can answer this questions better.

src/test/java/org/fogbowcloud/manager/core/plugins/cloud/openstack/OpenStackStateMapperTest.java
llvieira: I don't think this defensive programming is needed here, since we always pass a predefined `instanceType` and the `openStackState` will came from the cloud.

src/test/java/org/fogbowcloud/manager/core/plugins/cloud/openstack/OpenStackStateMapperTest.java
fubica: The code is tied to particular versions of each service (some are in V2 and others in V3), and the status codes can change from version to version. So, I suggested to Lucas to make the mapping function safer. The idea is to add a default clause that will set the state to INCONSISTENT whenever the state received from the cloud is not one that was anticipate when the plugins were implemented. This will make the code work even if the service version used in the cloud is not the same used to develop the plugin, and will signal the problem to the user. Lucas, please add also a LOGGER.error line in the default clauses, so that the site admin can quickly be informed of such problems.

llvieira: Okay, @fubica! :+1:

PR #83, opened 2018-07-13
Create tests and fix bug to orderController

llvieira:


src/test/java/org/fogbowcloud/manager/core/OrderControllerTest.java
```diff
@@ -101,6 +103,116 @@ public void testDeleteOrderStateClosed() throws UnexpectedException, OrderNotFou
...
+        FederationUser federationUser = new FederationUser("fake-id", null);
+        ComputeOrder computeOrder = new ComputeOrder();
+        computeOrder.setFederationUser(federationUser);
+        computeOrder.setRequestingMember(this.localMember);
+        computeOrder.setProvidingMember(this.localMember);
+        computeOrder.setOrderState(OrderState.OPEN);
+
+        this.openOrdersList.addItem(computeOrder);
+
+        this.ordersController.getAllOrders(federationUser, InstanceType.COMPUTE);
```
arn3tt: I think you were willing to use the return of this method call, right?

llvieira: Right!

PR #93, opened 2018-07-20
Refactory spawning processor test

wassoares:
- Removed static imports;

### DIFF
--- a/src/test/java/org/fogbowcloud/manager/core/processors/SpawningProcessorTest.java
+++ b/src/test/java/org/fogbowcloud/manager/core/processors/SpawningProcessorTest.java
@@ -20,15 +20,12 @@ import org.fogbowcloud.manager.util.connectivity.SshConnectivityUtil;
 import org.fogbowcloud.manager.util.connectivity.TunnelingServiceUtil;
 import org.junit.*;
 import org.junit.runner.RunWith;
+import org.mockito.BDDMockito;
 import org.mockito.Mockito;
 import org.powermock.api.mockito.PowerMockito;
 import org.powermock.core.classloader.annotations.PrepareForTest;
 import org.powermock.modules.junit4.PowerMockRunner;
 import java.util.HashMap;
-import static org.junit.Assert.*;
-import static org.mockito.BDDMockito.given;
-import static org.mockito.Matchers.anyString;
-import static org.mockito.Mockito.*;
 
 @RunWith(PowerMockRunner.class)
 @PrepareForTest(CloudConnectorFactory.class)
@@ -58,21 +55,22 @@ public class SpawningProcessorTest extends BaseUnitTests {
 
         HomeDir.getInstance().setPath(TEST_PATH);
 
-        this.tunnelingService = mock(TunnelingServiceUtil.class);
-        this.sshConnectivity = mock(SshConnectivityUtil.class);
-
-        LocalCloudConnector localCloudConnector = mock(LocalCloudConnector.class);
-
-        CloudConnectorFactory cloudConnectorFactory = mock(CloudConnectorFactory.class);
-        when(cloudConnectorFactory.getCloudConnector(anyString())).thenReturn(localCloudConnector);
+        CloudConnectorFactory cloudConnectorFactory = Mockito.mock(CloudConnectorFactory.class);
 
         PowerMockito.mockStatic(CloudConnectorFactory.class);
-        given(CloudConnectorFactory.getInstance()).willReturn(cloudConnectorFactory);
+        BDDMockito.given(CloudConnectorFactory.getInstance()).willReturn(cloudConnectorFactory);
+
+        LocalCloudConnector localCloudConnector = Mockito.mock(LocalCloudConnector.class);
+        Mockito.when(cloudConnectorFactory.getCloudConnector(Mockito.anyString()))
+                .thenReturn(localCloudConnector);
 
         this.cloudConnector = CloudConnectorFactory.getInstance()
                 .getCloudConnector(BaseUnitTests.LOCAL_MEMBER_ID);
 
-        this.spawningProcessor = spy(new SpawningProcessor(BaseUnitTests.LOCAL_MEMBER_ID,
+        this.tunnelingService = Mockito.mock(TunnelingServiceUtil.class);
+        this.sshConnectivity = Mockito.mock(SshConnectivityUtil.class);
+
+        this.spawningProcessor = Mockito.spy(new SpawningProcessor(BaseUnitTests.LOCAL_MEMBER_ID,
                 this.tunnelingService, this.sshConnectivity,
                 DefaultConfigurationConstants.SPAWNING_ORDERS_SLEEP_TIME));
 
@@ -93,11 +91,10 @@ public class SpawningProcessorTest extends BaseUnitTests {
         super.tearDown();
     }
 
-    // test case: If trying to process a ComputeOrder not spawning, the method
-    // processSpawningOrder() should not change OrderState to Fulfilled and
-    // should remain in its source list.
+    // test case: In calling the processSpawningOrder() method for any order other than spawning,
+    // you must not make state transition by keeping the order in your source list.
     @Test
-    public void testTryingProcessComputeOrderNotSpawning() throws Exception {
+    public void testProcessComputeOrderNotSpawning() throws Exception {
 
         // set up
         Order order = spyComputeOrder();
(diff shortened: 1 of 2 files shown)


src/test/java/org/fogbowcloud/manager/core/processors/SpawningProcessorTest.java
```diff
@@ -60,243 +55,279 @@ public void setUp() {
...
         order.setOrderState(OrderState.OPEN);
         this.openOrderList.addItem(order);
-        
+
+        // exercise
         this.spawningProcessor.processSpawningOrder(order);
 
-        assertNotNull(this.openOrderList.getNext());
-        assertNull(this.fulfilledOrderList.getNext());
+        // verify
```
thiagomanel: @wassoares I think it is better to check the order in the list is the one we expect instead of checking if there is one order

src/test/java/org/fogbowcloud/manager/core/processors/SpawningProcessorTest.java
```diff
@@ -60,243 +55,279 @@ public void setUp() {
...
-        
-        when(this.sshConnectivity.checkSSHConnectivity(any(SshTunnelConnectionData.class))).thenReturn(false);
-        
         this.thread = new Thread(this.spawningProcessor);
         this.thread.start();
         this.thread.sleep(500);
 
-        assertNotNull(this.spawningOrderList.getNext());
-        assertNull(this.fulfilledOrderList.getNext());
+        // verify
```
thiagomanel: same comment as above

src/test/java/org/fogbowcloud/manager/core/processors/SpawningProcessorTest.java
```diff
@@ -60,243 +55,279 @@ public void setUp() {
...
-        
+
         Instance orderInstance = Mockito.spy(new ComputeInstance(FAKE_INSTANCE_ID));
         orderInstance.setState(InstanceState.READY);
         order.setInstanceId(FAKE_INSTANCE_ID);
 
-        doReturn(orderInstance).when(this.cloudConnector).getInstance(any(Order.class));
+        // exercise
+        Mockito.doReturn(orderInstance).when(this.cloudConnector)
+                .getInstance(Mockito.any(Order.class));
```
thiagomanel: @wassoares we should not know these details in this test (getExternalServiceAddresses). Note that, the SpawningProcessor only knows getSshTunnelConnectionData and isInstanceReachable. Thus, the test is allowed to mock these two only.

src/test/java/org/fogbowcloud/manager/core/processors/SpawningProcessorTest.java
```diff
@@ -60,243 +55,279 @@ public void setUp() {
...
     }
-    
+
+    // test case: In calling the processSpawningOrder() method for any order other than spawning,
+    // you must not make state transition by keeping the order in your source list.
     @Test
-    public void testTryingProcessComputeOrderNotSpawning() throws Exception {
+    public void testProcessComputeOrderNotSpawning() throws Exception {
+
+        // set up
```
thiagomanel: @wassoares spies are really powerful. However, I'm little bit afraid of using them without an strong reason. Was not ok to use regular mocks instead or even the real objects?

src/test/java/org/fogbowcloud/manager/core/processors/SpawningProcessorTest.java
```diff
@@ -60,243 +55,279 @@ public void setUp() {
...
-    
+
+    // test case: When running thread in SpawningProcessor, if there is no SSH connectivity as the
+    // reverse tunnel, the method processSpawningOrder() must not change OrderState to Fulfilled and
+    // must remain in Spawning list.
     @SuppressWarnings("static-access")
     @Test
     public void testRunProcessComputeOrderWithoutSSHConnectivity() throws Exception {
+
+        // set up
```
thiagomanel: yeap ... I really do not like spies (it is one more thing to learn :))

src/test/java/org/fogbowcloud/manager/core/processors/SpawningProcessorTest.java
thiagomanel: ... I mean, it is better to check the element in the list is equals to the expected one.

src/test/java/org/fogbowcloud/manager/core/processors/SpawningProcessorTest.java
```diff
@@ -60,243 +55,279 @@ public void setUp() {
...
-    
+
+    // test case: When running thread in the SpawningProcessor and the OrderType is not a
+    // Compute, the processSpawningOrder() method must immediately change the OrderState to
+    // Fulfilled by adding in that list, and removed from the Spawning list.
     @SuppressWarnings("static-access")
     @Test
     public void testRunProcessWhenOrderTypeIsNotCompute() throws Exception {
+
+        // set up
```
thiagomanel: how about the other types? at least, add a fixme to remember we need to add the others.

src/test/java/org/fogbowcloud/manager/core/processors/SpawningProcessorTest.java
```diff
@@ -60,243 +55,279 @@ public void setUp() {
...
-        
-        Instance orderInstance = spy(new ComputeInstance(FAKE_INSTANCE_ID));
+        Assert.assertNull(this.fulfilledOrderList.getNext());
+
+        Instance orderInstance = Mockito.spy(new ComputeInstance(FAKE_INSTANCE_ID));
         orderInstance.setState(InstanceState.READY);
         order.setInstanceId(FAKE_INSTANCE_ID);
 
-        doReturn(orderInstance).when(this.cloudConnector).getInstance(any(Order.class));
+        // exercise
```
thiagomanel: doesn't this Mockito.doReturn be before the exercise block?

src/test/java/org/fogbowcloud/manager/core/processors/SpawningProcessorTest.java
```diff
@@ -60,243 +55,279 @@ public void setUp() {
...
-        Instance orderInstance = spy(new ComputeInstance(FAKE_INSTANCE_ID));
-        
+
+        Instance orderInstance = Mockito.spy(new ComputeInstance(FAKE_INSTANCE_ID));
         orderInstance.setState(InstanceState.DISPATCHED);
         order.setInstanceId(FAKE_INSTANCE_ID);
 
-        doReturn(orderInstance).when(this.cloudConnector).getInstance(Mockito.any(Order.class));
-        
+        // exercise
```
thiagomanel: it is not exercise, right?

src/test/java/org/fogbowcloud/manager/core/processors/SpawningProcessorTest.java
```diff
@@ -60,243 +55,279 @@ public void setUp() {
...
+        Mockito.doReturn(orderInstance).when(this.cloudConnector)
+                .getInstance(Mockito.any(Order.class));
+
         this.thread = new Thread(this.spawningProcessor);
         this.thread.start();
 
-        assertNotNull(this.spawningOrderList.getNext());
-        assertNull(this.fulfilledOrderList.getNext());
+        // verify
```
thiagomanel: maybe check equality?

src/test/java/org/fogbowcloud/manager/core/processors/SpawningProcessorTest.java
```diff
@@ -60,243 +55,279 @@ public void setUp() {
...
+        Instance orderInstance = Mockito.spy(new ComputeInstance(FAKE_INSTANCE_ID));
         orderInstance.setState(InstanceState.READY);
         order.setInstanceId(FAKE_INSTANCE_ID);
 
-        doReturn(orderInstance).when(this.cloudConnector).getInstance(any(Order.class));
-
-        when(this.tunnelingService.getExternalServiceAddresses(eq(order.getId()))).thenReturn(new HashMap<>());
-
-        when(this.sshConnectivity.checkSSHConnectivity(any(SshTunnelConnectionData.class))).thenReturn(true);
+        // exercise
```
thiagomanel: all these mock calls seem to be on the wrong place.

src/test/java/org/fogbowcloud/manager/core/processors/SpawningProcessorTest.java
```diff
@@ -60,243 +55,279 @@ public void setUp() {
...
-        Instance orderInstance = spy(new ComputeInstance(FAKE_INSTANCE_ID));
+        Instance orderInstance = Mockito.spy(new ComputeInstance(FAKE_INSTANCE_ID));
         orderInstance.setState(InstanceState.FAILED);
         order.setInstanceId(FAKE_INSTANCE_ID);
-        
-        doReturn(orderInstance).when(this.cloudConnector).getInstance(any(Order.class));
 
-        when(this.tunnelingService.getExternalServiceAddresses(eq(order.getId()))).thenReturn(new HashMap<>());
+        // exercise
+        Mockito.doReturn(orderInstance).when(this.cloudConnector)
```
thiagomanel: wrong place for mockito calls

PR #95, opened 2018-07-20
Refactoring tests in the 'OrderStateTransitionerTest' class

wassoares:
- Removed static imports;

### DIFF
--- a/src/test/java/org/fogbowcloud/manager/OrderStateTransitionerTest.java
+++ b/src/test/java/org/fogbowcloud/manager/OrderStateTransitionerTest.java
@@ -27,12 +27,6 @@ public class OrderStateTransitionerTest extends BaseUnitTests {
 
     private MockUtil mockUtil = new MockUtil();
 
-    private Order createOrder(OrderState orderState) {
-        Order order = createLocalOrder(getLocalMemberId());
-        order.setOrderState(orderState);
-        return order;
-    }
-
     @Before
     public void setUp() {
         HomeDir.getInstance().setPath("src/test/resources/private");
@@ -42,27 +36,24 @@ public class OrderStateTransitionerTest extends BaseUnitTests {
     public void tearDown() {
         SharedOrderHolders instance = SharedOrderHolders.getInstance();
 
-        /** Clearing the lists if we are not mocking them */
+        // Clearing the lists if we are not mocking them
         if (!mockUtil.isMock(instance)) {
-            for (OrderState state : OrderState.values()) {
-                if (!state.equals(OrderState.DEACTIVATED)) {
-                    SynchronizedDoublyLinkedList ordersList = instance.getOrdersList(state);
-
-                    ordersList.resetPointer();
-                    Order order;
-                    while ((order = ordersList.getNext()) != null) {
-                        ordersList.removeItem(order);
-                    }
-                }
-            }
+            super.tearDown();
         }
     }
 
-    // test case: When calling the transition() method passing an Order and a OrderState, this
-    // order will be change to the specific OrderState, removing of the origin list, and adding in
-    // that state list.
+    private Order createOrder(OrderState orderState) {
+        Order order = createLocalOrder(getLocalMemberId());
+        order.setOrderState(orderState);
+        return order;
+    }
+
+    // test case: When calling the transition() method, it must change the state of Order
+    // passed per parameter to a specific OrderState. This Order defined originally with Open, will
+    // be change to the Spawning state, removing of the open orders list, and adding in the spawning
+    // orders list.
     @Test
-    public void testValidTransition() throws UnexpectedException {
+    public void testTransitionToChangeOrderStateOpenToSpawning() throws UnexpectedException {
 
         // set up
         OrderState originState = OrderState.OPEN;
@@ -92,13 +83,21 @@ public class OrderStateTransitionerTest extends BaseUnitTests {
 
         Order order = createOrder(originState);
         openOrdersList.addItem(order);
+        Assert.assertNull(spawningOrdersList.getNext());
 
         // exercise
         OrderStateTransitioner.transition(order, destinationState);
 
         // verify
-        Assert.assertNull(openOrdersList.getNext());
+        Mockito.verify(databaseManager, Mockito.times(1)).readActiveOrders(OrderState.OPEN);
+        Mockito.verify(databaseManager, Mockito.times(1)).readActiveOrders(OrderState.SPAWNING);
+        Mockito.verify(databaseManager, Mockito.times(1)).readActiveOrders(OrderState.FAILED);
+        Mockito.verify(databaseManager, Mockito.times(1)).readActiveOrders(OrderState.FULFILLED);
+        Mockito.verify(databaseManager, Mockito.times(1)).readActiveOrders(OrderState.PENDING);
+        Mockito.verify(databaseManager, Mockito.times(1)).readActiveOrders(OrderState.CLOSED);
+
         Assert.assertEquals(order, spawningOrdersList.getNext());
+        Assert.assertNull(openOrdersList.getNext());
     }
 
     // test case: When calling the transition() method and the origin list of the 'Order' is Null,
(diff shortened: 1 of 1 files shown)


src/test/java/org/fogbowcloud/manager/OrderStateTransitionerTest.java
```diff
@@ -34,7 +32,7 @@ private Order createOrder(OrderState orderState) {
         order.setOrderState(orderState);
```
thiagomanel: move this createOrder method to after the tearDown

src/test/java/org/fogbowcloud/manager/OrderStateTransitionerTest.java
```diff
@@ -44,10 +42,10 @@ public void setUp() {
     public void tearDown() {
         SharedOrderHolders instance = SharedOrderHolders.getInstance();
```
thiagomanel: avoid /** comments within methods. /** is for javadoc

src/test/java/org/fogbowcloud/manager/OrderStateTransitionerTest.java
```diff
@@ -60,21 +58,32 @@ public void tearDown() {
         }
     }
 
+    // test case: When calling the transition() method passing an Order and a OrderState, this
+    // order will be change to the specific OrderState, removing of the origin list, and adding in
+    // that state list.
     @Test
     public void testValidTransition() throws UnexpectedException {
```
thiagomanel: note that we test only the open to spawning transition. I suggest to change the name of the method to better reflect what it does.

thiagomanel: ... change the description too.

src/test/java/org/fogbowcloud/manager/OrderStateTransitionerTest.java
```diff
@@ -84,73 +93,95 @@ public void testValidTransition() throws UnexpectedException {
         Order order = createOrder(originState);
         openOrdersList.addItem(order);
 
+        // exercise
         OrderStateTransitioner.transition(order, destinationState);
-        assertNull(openOrdersList.getNext());
-        assertEquals(order, spawningOrdersList.getNext());
+
+        // verify
```
thiagomanel: it would be nice to assert is was different before the exercise

src/test/java/org/fogbowcloud/manager/OrderStateTransitionerTest.java
```diff
@@ -84,73 +93,95 @@ public void testValidTransition() throws UnexpectedException {
...
+    // will be throws an unexpected exception.
+    @Test(expected = UnexpectedException.class) // verify
     public void testOriginListCannotBeFound() throws UnexpectedException {
+
+        // set up
         OrderState originState = OrderState.OPEN;
         OrderState destinationState = OrderState.SPAWNING;
 
-        // origin list will fail to be found
+        /** Origin list will fail to be found */
```
thiagomanel: avoid javadoc here

src/test/java/org/fogbowcloud/manager/OrderStateTransitionerTest.java
```diff
@@ -84,73 +93,95 @@ public void testValidTransition() throws UnexpectedException {
         Order order = createOrder(originState);
         openOrdersList.addItem(order);
 
+        // exercise
         OrderStateTransitioner.transition(order, destinationState);
-        assertNull(openOrdersList.getNext());
-        assertEquals(order, spawningOrdersList.getNext());
+
+        // verify
+        Assert.assertNull(openOrdersList.getNext());
```
arn3tt: Don't we need to verify the calls to the databaseManager mock?

src/test/java/org/fogbowcloud/manager/OrderStateTransitionerTest.java
```diff
@@ -84,73 +93,95 @@ public void testValidTransition() throws UnexpectedException {
...
+
         try {
+
+            // exercise
             OrderStateTransitioner.transition(order, destinationState);
         } catch (UnexpectedException e) {
-            fail();
+
+            // verify
+            Assert.fail();
```
arn3tt: This doesn't looks like a verification.

src/test/java/org/fogbowcloud/manager/OrderStateTransitionerTest.java
```diff
@@ -84,73 +93,95 @@ public void testValidTransition() throws UnexpectedException {
...
         }
```
arn3tt: Shouldn't we verify the calls to the stub objects?

src/test/java/org/fogbowcloud/manager/OrderStateTransitionerTest.java
```diff
@@ -84,73 +93,95 @@ public void testValidTransition() throws UnexpectedException {
...
 
-        
         Order order = createOrder(originState);
-        
+
         try {
+
+            // exercise
             OrderStateTransitioner.transition(order, destinationState);
         } catch (UnexpectedException e) {
```
thiagomanel: it seems we do not need try/catch construct here. Don't we test the same thing if we remove it (note that, if there is a bug in the code we are testing, an exception will be throw and the test will fail)

wassoares: Test was removed because it no longer has its initial function.

src/test/java/org/fogbowcloud/manager/OrderStateTransitionerTest.java
wassoares: Ok, I get it ;)

src/test/java/org/fogbowcloud/manager/OrderStateTransitionerTest.java
wassoares: Fixed.

src/test/java/org/fogbowcloud/manager/OrderStateTransitionerTest.java
wassoares: Done.

src/test/java/org/fogbowcloud/manager/OrderStateTransitionerTest.java
wassoares: Ok, done.

src/test/java/org/fogbowcloud/manager/OrderStateTransitionerTest.java
wassoares: Done!

src/test/java/org/fogbowcloud/manager/OrderStateTransitionerTest.java
wassoares: Test was removed because it no longer has its initial function.

src/test/java/org/fogbowcloud/manager/OrderStateTransitionerTest.java
wassoares: Test was removed because it no longer has its initial function.

src/test/java/org/fogbowcloud/manager/OrderStateTransitionerTest.java
wassoares: Not because currently this database script has been replaced by the method call 'mockReadOrdersFromDataBase()' of the super class 'BaseUnitTests'.

src/test/java/org/fogbowcloud/manager/OrderStateTransitionerTest.java
```diff
@@ -44,10 +42,10 @@ public void setUp() {
     public void tearDown() {
         SharedOrderHolders instance = SharedOrderHolders.getInstance();
 
-        // Clearing the lists if we are not mocking them
+        /** Clearing the lists if we are not mocking them */
         if (!mockUtil.isMock(instance)) {
             for (OrderState state : OrderState.values()) {
```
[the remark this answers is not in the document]
wassoares: Fixed.

PR #96, opened 2018-07-20
AaControllerTest according to the new template definitions

ordansantos:


src/test/java/org/fogbowcloud/manager/core/AaControllerTest.java
```diff
@@ -48,112 +48,174 @@ public void setUp() {
     	
     }
     
+    //test case: Check if authenticate method throws no exception when the federation token is valid.
```
thiagomanel: lines 41 and 47 are duplicated. Could you check this?

src/test/java/org/fogbowcloud/manager/core/AaControllerTest.java
```diff
@@ -48,112 +48,174 @@ public void setUp() {
...
     @Test
     public void testAuthenticate() throws UnauthenticatedUserException {
+    	//set up
```
thiagomanel: In addition to that, I'd make sure that the same string passed as arg in this.asController.authenticate is passed to the federationIdentityPlugin (so, add a verify to this test case)

src/test/java/org/fogbowcloud/manager/core/AaControllerTest.java
```diff
@@ -48,112 +48,174 @@ public void setUp() {
...
     @Test (expected = UnauthenticatedUserException.class)
     public void testAuthenticateWhenUnauthenticatedUserException() throws UnauthenticatedUserException {
+    	//set up
     	Mockito.when(this.federationIdentityPluginMock.isValid(Mockito.anyString())).thenReturn(false);
+    	
+    	//exercise/verify
     	this.aaController.authenticate(Mockito.anyString());
     }
     
+    //test case: Check if authenticate method throws no exception when the operation is valid.
```
thiagomanel: typo. authorize instead of authenticate in line 71, right?

thiagomanel: below method also have typos

src/test/java/org/fogbowcloud/manager/core/AaControllerTest.java
```diff
@@ -48,112 +48,174 @@ public void setUp() {
...
+    	//set up
     	Mockito.when(this.authorizationPluginMock.isAuthorized(
     			Mockito.any(), 
     			Mockito.any())).thenReturn(true);
+    	
+    	//exercise/verify
     	this.aaController.authorize(Mockito.any(), Mockito.any());
     }
     
+    //test case: Check if authenticate method throws Unauthenticated exception when the federation token is valid.
```
thiagomanel: "... token is valid" or "token is invalid"? check the order test cases.

src/test/java/org/fogbowcloud/manager/core/AaControllerTest.java
```diff
@@ -48,112 +48,174 @@ public void setUp() {
...
     	String accessId = "accessId";
     	User user = new User("id", "name");
     	Date date = new Date();
     	Map <String, String> attributes = new HashMap<String, String>();
-    	Token token = new Token(accessId, user, date, attributes);
-    	Mockito.when(this.localIdentityPluginMock.createToken(Mockito.any())).thenReturn(token);
-    	Assert.assertEquals(token, this.aaController.getLocalToken(null));
+    	Token expectedToken = new Token(accessId, user, date, attributes);
+    	Mockito.when(this.localIdentityPluginMock.createToken(Mockito.any())).thenReturn(expectedToken);
+
```
thiagomanel: although the test passed when the user is null, it is misleading (the very first it is that you are testing the error conditions because of null references. I'd replace by a properly constructed value.

src/test/java/org/fogbowcloud/manager/core/AaControllerTest.java
```diff
@@ -48,112 +48,174 @@ public void setUp() {
...
+    	Mockito.when(this.localIdentityPluginMock.createToken(Mockito.any())).thenReturn(expectedToken);
+    	
+    	//exercise
+    	Token aaControllerToken = this.aaController.getLocalToken(null);
+    	
+    	//verify
+    	Assert.assertEquals(expectedToken, aaControllerToken);
     }
     
+    //test case: Check if getLocalToken is returning FogbowManagerException from LocalIdentityPlugin properly.
```
thiagomanel: a better doc would say something like "... is properly forwarding the exception thrown by ..."

ordansantos: Much better described, I'll use this template now, thank you

src/test/java/org/fogbowcloud/manager/core/AaControllerTest.java
ordansantos: copy-paste trouble

src/test/java/org/fogbowcloud/manager/core/AaControllerTest.java
ordansantos: copy-paste trouble

src/test/java/org/fogbowcloud/manager/core/AaControllerTest.java
ordansantos: Adding verify is really a good way to improve this type of test, thank you for the observation

PR #101, opened 2018-07-23
Improved SharedOrderHoldersTest according the new template definitions

llvieira:


src/test/java/org/fogbowcloud/manager/SharedOrderHoldersTest.java
```diff
@@ -27,39 +21,42 @@
...
+    // they are the same object.
     @Test
     public void testGetSameListReference() {
+        // set up
         SynchronizedDoublyLinkedList listFromInstanceOne = instanceOne.getOpenOrdersList();
         SynchronizedDoublyLinkedList listFromInstanceTwo = instanceTwo.getOpenOrdersList();
-        assertEquals(listFromInstanceOne, listFromInstanceTwo);
 
+        // verify
+        Assert.assertEquals(listFromInstanceOne, listFromInstanceTwo);
```
arn3tt: Should we test this using double equals signs (==) instead of equals? See @[this post](https://stackoverflow.com/questions/7520432/what-is-the-difference-between-vs-equals-in-java) for more info

src/test/java/org/fogbowcloud/manager/SharedOrderHoldersTest.java
```diff
@@ -27,39 +21,42 @@
...
+        Assert.assertEquals(listFromInstanceOne.getCurrent(), listFromInstanceTwo.getCurrent());
+        Assert.assertEquals(orderOne, listFromInstanceOne.getCurrent().getOrder());
+        Assert.assertEquals(orderOne, listFromInstanceTwo.getCurrent().getOrder());
+
+        // exercise
         Order orderTwo = createLocalOrder(getLocalMemberId());
         listFromInstanceTwo.addItem(orderTwo);
-        assertEquals(
+
+        // verify
```
arn3tt: Again, I think the verify keyword should be used when verifying mock objects.

src/test/java/org/fogbowcloud/manager/SharedOrderHoldersTest.java
llvieira: In Junit we have assertSame (same of ==) which assert that two objects refer to the same object. I will fix it.

PR #106, opened 2018-07-23
Refactoring tests in the 'ApplicationFacadeTest' class

wassoares:
- Removed static imports;

### DIFF
--- a/src/test/java/org/fogbowcloud/manager/core/ApplicationFacadeTest.java
+++ b/src/test/java/org/fogbowcloud/manager/core/ApplicationFacadeTest.java
@@ -12,1057 +12,1367 @@ import org.junit.Assert;
 import org.junit.Before;
 import org.junit.Test;
 import org.junit.runner.RunWith;
+import org.mockito.BDDMockito;
+import org.mockito.Matchers;
 import org.mockito.Mockito;
 import org.powermock.api.mockito.PowerMockito;
 import org.powermock.core.classloader.annotations.PrepareForTest;
 import org.powermock.modules.junit4.PowerMockRunner;
-
 import java.util.ArrayList;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
 
-import static org.mockito.BDDMockito.given;
-import static org.mockito.Matchers.any;
-import static org.mockito.Mockito.doNothing;
-import static org.mockito.Mockito.when;
-
 @RunWith(PowerMockRunner.class)
 @PrepareForTest(DatabaseManager.class)
 public class ApplicationFacadeTest extends BaseUnitTests {
 
-	private ApplicationFacade application;
-	private AaController aaaController;
-	private OrderController orderController;
+    private static final String FAKE_ORDER_ID = "fake-order-id";
+    private static final String FAKE_INSTANCE_ID = "fake-instance-id";
+    private static final String FEDERATION_TOKEN_VALUE = "federation_token-value";
+    private static final String FAKE_USER = "fake-user";
+    private static final String FAKE_MEMBER_ID = "fake-member-id";
+    private static final String FAKE_GATEWAY = "fake-gateway";
+    private static final String FAKE_ADDRESS = "fake-address";
+    private static final String FAKE_VOLUME_NAME = "fake-volume-name";
+    private static final String FAKE_IMAGE_NAME = "fake-image-name";
+    private static final String FAKE_PUBLIC_KEY = "fake-public-key";
+    private static final String FAKE_SOURCE_ID = "fake-source-id";
+    private static final String FAKE_TARGET_ID = "fake-target-id";
+    private static final String FAKE_DEVICE_MOUNT_POINT = "fake-device-mount-point";
+
+    private ApplicationFacade application;
+    private AaController aaaController;
+    private OrderController orderController;
     private Map<String, Order> activeOrdersMap;
 
-	@Before
-	public void setUp() throws UnauthorizedRequestException {
-		this.aaaController = Mockito.mock(AaController.class);
-		
-		HomeDir.getInstance().setPath("src/test/resources/private");
+    @Before
+    public void setUp() throws UnauthorizedRequestException {
+        this.aaaController = Mockito.mock(AaController.class);
 
-		DatabaseManager databaseManager = Mockito.mock(DatabaseManager.class);
-		when(databaseManager.readActiveOrders(OrderState.OPEN)).thenReturn(new SynchronizedDoublyLinkedList());
-		when(databaseManager.readActiveOrders(OrderState.SPAWNING)).thenReturn(new SynchronizedDoublyLinkedList());
-		when(databaseManager.readActiveOrders(OrderState.FAILED)).thenReturn(new SynchronizedDoublyLinkedList());
-		when(databaseManager.readActiveOrders(OrderState.FULFILLED)).thenReturn(new SynchronizedDoublyLinkedList());
-		when(databaseManager.readActiveOrders(OrderState.PENDING)).thenReturn(new SynchronizedDoublyLinkedList());
-		when(databaseManager.readActiveOrders(OrderState.CLOSED)).thenReturn(new SynchronizedDoublyLinkedList());
+        HomeDir.getInstance().setPath("src/test/resources/private");
 
-		doNothing().when(databaseManager).add(any(Order.class));
-		doNothing().when(databaseManager).update(any(Order.class));
+        DatabaseManager databaseManager = Mockito.mock(DatabaseManager.class);
+        Mockito.when(databaseManager.readActiveOrders(OrderState.OPEN))
+                .thenReturn(new SynchronizedDoublyLinkedList());
+        Mockito.when(databaseManager.readActiveOrders(OrderState.SPAWNING))
+                .thenReturn(new SynchronizedDoublyLinkedList());
+        Mockito.when(databaseManager.readActiveOrders(OrderState.FAILED))
+                .thenReturn(new SynchronizedDoublyLinkedList());
+        Mockito.when(databaseManager.readActiveOrders(OrderState.FULFILLED))
+                .thenReturn(new SynchronizedDoublyLinkedList());
+        Mockito.when(databaseManager.readActiveOrders(OrderState.PENDING))
+                .thenReturn(new SynchronizedDoublyLinkedList());
+        Mockito.when(databaseManager.readActiveOrders(OrderState.CLOSED))
+                .thenReturn(new SynchronizedDoublyLinkedList());
 
-		PowerMockito.mockStatic(DatabaseManager.class);
-		given(DatabaseManager.getInstance()).willReturn(databaseManager);
-		
-		this.orderController = Mockito.spy(new OrderController());
-		this.application = Mockito.spy(ApplicationFacade.getInstance());
-		this.application.setAaController(this.aaaController);
-		this.application.setOrderController(this.orderController);
+        Mockito.doNothing().when(databaseManager).add(Matchers.any(Order.class));
+        Mockito.doNothing().when(databaseManager).update(Matchers.any(Order.class));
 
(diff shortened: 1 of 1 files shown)


src/test/java/org/fogbowcloud/manager/core/ApplicationFacadeTest.java
```diff
@@ -12,1701 +12,2209 @@
...
+                .thenReturn(new SynchronizedDoublyLinkedList());
+
+        Mockito.doNothing().when(databaseManager).add(Matchers.any(Order.class));
+        Mockito.doNothing().when(databaseManager).update(Matchers.any(Order.class));
+
+        PowerMockito.mockStatic(DatabaseManager.class);
+        BDDMockito.given(DatabaseManager.getInstance()).willReturn(databaseManager);
+
+        this.orderController = Mockito.spy(new OrderController());
+        this.application = Mockito.spy(ApplicationFacade.getInstance());
```
thiagomanel: Please, check if it we really need to use a spies. If not, avoid using them.

src/test/java/org/fogbowcloud/manager/core/ApplicationFacadeTest.java
```diff
@@ -12,1701 +12,2209 @@
...
-        expectedList.add("fake-order-id");
+        expectedList.add(FAKE_ORDER_ID);
 
         ComputeOrder computeOrder = Mockito.mock(ComputeOrder.class);
         Mockito.doReturn(expectedList).when(computeOrder).getNetworksId();
 
+        // exercise
         this.application.changeNetworkOrderIdsToNetworInstanceIds(computeOrder);
 
+        // verify
```
thiagomanel: I think we cannot mock the return of the computerOrder.getNetworkId. This is the behaviour we want to check. One option is to not use a mock for it or to use and to check we called the set method

src/test/java/org/fogbowcloud/manager/core/ApplicationFacadeTest.java
```diff
@@ -12,1701 +12,2209 @@
...
-		Order order = createComputeOrder();
-		OrderStateTransitioner.activateOrder(order);
+        // verify
+        Assert.assertEquals(OrderState.CLOSED, order.getOrderState());
+    }
+
+    // test case: When try calling the deleteCompute() method without user authentication, it must
+    // throw UnauthenticatedUserException and the Order remains in the same state.
+    @Test
+    public void testDeleteComputeOrderUnathenticated() throws Exception {
```
thiagomanel: I think we can improve the name of this test. It seems we are testing the forward of exception

thiagomanel: ... to me, it was hard to understand because the name has "Unathenticated". We are doing authorization, in fact.

Also, there is a typo in the name.

thiagomanel: ... in addition to that, the method do no call authenticate. Which indicates we are missing verifies in the code.

src/test/java/org/fogbowcloud/manager/core/ApplicationFacadeTest.java
```diff
@@ -12,1701 +12,2209 @@
...
+        Mockito.doReturn(order).when(this.orderController).getOrder(Mockito.anyString(),
+                Mockito.any(FederationUser.class), Mockito.any(InstanceType.class));
 
-		Mockito.doNothing().when(this.aaaController).authorize(Mockito.any(FederationUser.class),
-				Mockito.any(Operation.class), Mockito.any(Order.class));
+        Mockito.doNothing().when(this.aaaController).authorize(Mockito.any(FederationUser.class),
+                Mockito.any(Operation.class), Mockito.any(Order.class));
+
+        Mockito.doThrow(new UnauthenticatedUserException()).when(this.aaaController)
+                .authenticate(Mockito.anyString());
```
thiagomanel: why mocking this, if we do not call authenticate in the method under test?

src/test/java/org/fogbowcloud/manager/core/ApplicationFacadeTest.java
```diff
@@ -12,1701 +12,2209 @@
...
 
-		try {
-			String federationTokenValue = "";
-			this.application.deleteCompute(order.getId(), federationTokenValue);
-			Assert.fail();
-		} catch (UnauthenticatedUserException e) {
-			Assert.assertEquals(OrderState.OPEN, order.getOrderState());
-		}
-	}
+    // test case: When try calling the deleteCompute() method without a token authentication, it
```
thiagomanel: I think we should improve the description and the code. It seems to me we are testing the facade behaviour when the controller.getFederationUse throws an exception. This is unrelated to "without a token authentication"

src/test/java/org/fogbowcloud/manager/core/ApplicationFacadeTest.java
```diff
@@ -12,1701 +12,2209 @@
...
+            // verify
+            Assert.assertEquals(OrderState.OPEN, order.getOrderState());
+        }
+    }
 
-		Order order = createComputeOrder();
-		OrderStateTransitioner.activateOrder(order);
+    // test case: When try calling the deleteCompute() method performing an operation without
+    // authorization, it must throw UnauthorizedRequestException and the Order remains in the same
+    // state.
```
thiagomanel: improve the description of the test case to reflect what the test do (or change the test to address the description). see my previous comments

src/test/java/org/fogbowcloud/manager/core/ApplicationFacadeTest.java
```diff
@@ -12,1701 +12,2209 @@
...
+            this.application.deleteCompute(order.getId(), FEDERATION_TOKEN_VALUE);
+            Assert.fail();
+        } catch (UnauthorizedRequestException e) {
+            // verify
+            Assert.assertEquals(OrderState.OPEN, order.getOrderState());
+        }
+    }
 
-		Mockito.doReturn(order.getFederationUser()).when(this.aaaController).getFederationUser(Mockito.anyString());
+    // test case: When calling the createCompute() method the Order passed by parameter must return
```
thiagomanel: I did not understand the case. Are we trying to check the order state was correctly assigned after the exercise? If so, what was the state before the exercise?

src/test/java/org/fogbowcloud/manager/core/ApplicationFacadeTest.java
```diff
@@ -12,1701 +12,2209 @@
...
-		}
-	}
+        Mockito.doThrow(new UnauthenticatedUserException()).when(this.aaaController)
+                .authenticate(Mockito.anyString());
 
-	@Test
-	public void testCreateComputeOrderTokenUnauthenticated() throws Exception {
-		ComputeOrder order = createComputeOrder();
+        Mockito.doReturn(order.getFederationUser()).when(this.aaaController)
+                .getFederationUser(Mockito.anyString());
```
thiagomanel: I think there is something wrong in the mock setup. The point is: how can this order.getFederationUser be called if we are going to throw an exception earlier? This is a symptom we are missing verifies. Also the description of the test case is not code, see the comment I made before.

src/test/java/org/fogbowcloud/manager/core/ApplicationFacadeTest.java
```diff
@@ -12,1701 +12,2209 @@
...
 
-		try {
-			String federationTokenValue = "";
-			this.application.createCompute(order, federationTokenValue);
-			Assert.fail();
-		} catch (UnauthenticatedUserException e) {
-			Assert.assertNull(order.getOrderState());
-		}
-	}
+        Mockito.doThrow(new UnauthenticatedUserException()).when(this.aaaController)
```
thiagomanel: same problems as before

src/test/java/org/fogbowcloud/manager/core/ApplicationFacadeTest.java
```diff
@@ -12,1701 +12,2209 @@
...
+            Assert.fail();
+        } catch (UnauthenticatedUserException e) {
+            // verify
+            Assert.assertNull(order.getOrderState());
+        }
+    }
 
-		Mockito.doNothing().when(this.aaaController).authorize(Mockito.any(FederationUser.class), Mockito.any(Operation.class),
-				Mockito.any(Order.class));
+    // test case: When try calling the createCompute() method with a Federation User without
```
thiagomanel: same problems as before

src/test/java/org/fogbowcloud/manager/core/ApplicationFacadeTest.java
wassoares: The spy on line 76 can be removed without affecting the code.
however, at line 75 its removal would cause failure in 12 of the 81 tests, and then I'll keep it.

src/test/java/org/fogbowcloud/manager/core/ApplicationFacadeTest.java
wassoares: Will be fixed at the next pull request...

src/test/java/org/fogbowcloud/manager/core/ApplicationFacadeTest.java
```diff
@@ -12,1701 +12,2209 @@
...
 
-	@Test
-	public void testDeleteComputeOrderUnathenticated() throws Exception {
-		Mockito.doThrow(new UnauthenticatedUserException()).when(this.aaaController).authenticate(Mockito.anyString());
+        // exercise
+        this.application.deleteCompute(order.getId(), FEDERATION_TOKEN_VALUE);
 
-		Order order = createComputeOrder();
-		OrderStateTransitioner.activateOrder(order);
+        // verify
```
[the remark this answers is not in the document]
wassoares: Will be fixed at the next pull request...

src/test/java/org/fogbowcloud/manager/core/ApplicationFacadeTest.java
wassoares: Modified the test method name for testDeleteComputeOrderWithoutAuthentication().

src/test/java/org/fogbowcloud/manager/core/ApplicationFacadeTest.java
wassoares: The 'deleteCompute()' method calls 'deleteOrder()' from the 'ApplicationFacade' class. This method 'deleteOrder()' before delegating its effective action to remove the order, requires the AaController class to authenticate the 'federationTokenValue'. And apparently 'Mock' is used to catch the exception thrown in the 'authenticate()' method of this class... 
But for me, this should only be tested in the 'AaControllerTest' class.

src/test/java/org/fogbowcloud/manager/core/ApplicationFacadeTest.java
wassoares: Fixed the name and description of the test case.

src/test/java/org/fogbowcloud/manager/core/ApplicationFacadeTest.java
wassoares: Yes, the previous 'OrderState' was 'null'. Anyway, I fixed the description of the test case.

src/test/java/org/fogbowcloud/manager/core/ApplicationFacadeTest.java
wassoares: I will try to modify.

src/test/java/org/fogbowcloud/manager/core/ApplicationFacadeTest.java
wassoares: I will try to modify.

src/test/java/org/fogbowcloud/manager/core/ApplicationFacadeTest.java
wassoares: I will try to modify.

PR #107, opened 2018-07-23
Enhancing OpenStackStateMapper test according new definitions

llvieira:
I don't think the comment `// test case: ...` is needed here.

### DIFF
--- a/src/main/java/org/fogbowcloud/manager/core/plugins/cloud/openstack/OpenStackStateMapper.java
+++ b/src/main/java/org/fogbowcloud/manager/core/plugins/cloud/openstack/OpenStackStateMapper.java
@@ -8,28 +8,28 @@ public class OpenStackStateMapper {
 
     private static final Logger LOGGER = Logger.getLogger(OpenStackStateMapper.class);
 
-    public static final String ACTIVE_STATUS = "active";
-    public static final String BUILD_STATUS = "build";
-    public static final String DOWN_STATUS = "down";
-    public static final String ERROR_STATUS = "error";
-    public static final String CREATING_STATUS = "creating";
-    public static final String AVAILABLE_STATUS = "available";
-    public static final String ATTACHING_STATUS = "attaching";
-    public static final String DETACHING_STATUS = "detaching";
-    public static final String IN_USE_STATUS = "in-use";
-    public static final String MAINTENANCE_STATUS = "maintenance";
-    public static final String DELETING_STATUS = "deleting";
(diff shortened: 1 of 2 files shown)


src/main/java/org/fogbowcloud/manager/core/plugins/cloud/openstack/OpenStackStateMapper.java
```diff
@@ -12,24 +12,24 @@
...
-    private static final String BACKING_UP_STATUS = "backing-up";
-    private static final String RESTORING_BACKUP_STATUS = "restoring-backup";
-    private static final String ERROR_BACKING_UP_STATUS = "error_backing-up";
-    private static final String ERROR_RESTORING_STATUS = "error_restoring";
-    private static final String ERROR_EXTENDING_STATUS = "error_extending";
-    private static final String DOWNLOADING_STATUS = "downloading";
-    private static final String UPLOADING_STATUS = "uploading";
-    private static final String RETYPING_STATUS = "retyping";
-    private static final String EXTENDING_STATUS = "extending";
+    public static final String CREATING_STATUS = "creating";
```
thiagomanel: @llvieira I'm not sure we need this to be public. If you are using only for tests, shouldn't be protected?

src/test/java/org/fogbowcloud/manager/core/plugins/cloud/openstack/OpenStackStateMapperTest.java
```diff
@@ -9,342 +9,383 @@
 
     @Test
     public void testComputeActiveStatusToReady() {
+        // set up
         InstanceType instanceType = InstanceType.COMPUTE;
 
-        InstanceState instanceState1 = OpenStackStateMapper.map(instanceType, "active");
-        InstanceState instanceState2 = OpenStackStateMapper.map(instanceType, "ACTIVE");
+        // exercise
```
thiagomanel: I think you are swapping, again, actual and expected values ....

src/main/java/org/fogbowcloud/manager/core/plugins/cloud/openstack/OpenStackStateMapper.java
llvieira: I made it protected now.

src/test/java/org/fogbowcloud/manager/core/plugins/cloud/openstack/OpenStackStateMapperTest.java
llvieira: Fixed!

PR #108, opened 2018-07-23
Enhancing OpenStackComputeQuotaPlugin tests according to the new specifications

llvieira:


src/test/java/org/fogbowcloud/manager/core/plugins/cloud/openstack/OpenStackComputeQuotaPluginTest.java
```diff
@@ -40,41 +40,56 @@ public void setUp() {
...
 	    Mockito.doThrow(FogbowManagerException.class).when(this.plugin).getUserQuota(this.token);
+
+	    // verify
 	    Mockito.verify(this.plugin).getUserQuota(Mockito.eq(this.token));
 
 	}
-	
-	@Test (expected = FogbowManagerException.class)
+
+    // test case: Verify getUserQuota() when it raises UnexpectedException.
```
ordansantos: This method is actually doing nothing, please, run code coverage and notice this method breaks in the very first line.

src/test/java/org/fogbowcloud/manager/core/plugins/cloud/openstack/OpenStackComputeQuotaPluginTest.java
```diff
@@ -40,41 +40,56 @@ public void setUp() {
...
-        Assert.assertEquals(allocationAvailable.getvCPU(), this.plugin.getUserQuota(this.token).getAvailableQuota().getvCPU());
-        Assert.assertEquals(allocationAvailable.getRam(), this.plugin.getUserQuota(this.token).getAvailableQuota().getRam());
-        Assert.assertEquals(allocationAvailable.getInstances(), this.plugin.getUserQuota(this.token).getAvailableQuota().getInstances());
+        Assert.assertEquals(allocationAvailable.getvCPU(), availableQuota.getvCPU());
+        Assert.assertEquals(allocationAvailable.getRam(), availableQuota.getRam());
+        Assert.assertEquals(allocationAvailable.getInstances(), availableQuota.getInstances());
     }
-	
+
+    // test case: Verify getUserQuota() when it raises FogbowManagerException.
```
ordansantos: This method is actually doing nothing, please, run code coverage and notice this method breaks in the very first line.

ordansantos: In addition to that, it seems this method is trying to test a mocked method, I think we should replace this whole test with another one.

src/test/java/org/fogbowcloud/manager/core/plugins/cloud/openstack/OpenStackComputeQuotaPluginTest.java
```diff
@@ -40,41 +40,56 @@ public void setUp() {
 		token.setAttributes(attributes);
 		token.setAccessId("");
 	}
-	
+
+	// test case: Tests if getTotalQuota(), getUsedQuota() and getAvailableQuota() returns the right quotas.
     @Test
     public void getUserQuotaTest() throws FogbowManagerException, UnexpectedException {
+	    // set up
         ComputeAllocation allocationTotal = new ComputeAllocation(64, 46080, 20);
```
ordansantos: Where these magic numbers come from? Were they calculated by hand? We should think in a better way to generate this type of data.

src/test/java/org/fogbowcloud/manager/core/plugins/cloud/openstack/OpenStackComputeQuotaPluginTest.java
```diff
@@ -40,41 +40,56 @@ public void setUp() {
 		token.setAttributes(attributes);
 		token.setAccessId("");
 	}
-	
+
+	// test case: Tests if getTotalQuota(), getUsedQuota() and getAvailableQuota() returns the right quotas.
```
ordansantos: returns the right quotas from where?

src/test/java/org/fogbowcloud/manager/core/plugins/cloud/openstack/OpenStackComputeQuotaPluginTest.java
llvieira: I agree, this whole test make no sense. I'm going to take it away.

src/test/java/org/fogbowcloud/manager/core/plugins/cloud/openstack/OpenStackComputeQuotaPluginTest.java
llvieira: Another one which does not make sense.

src/test/java/org/fogbowcloud/manager/core/plugins/cloud/openstack/OpenStackComputeQuotaPluginTest.java
llvieira: Now I reproduced the method where it is generated.

PR #114, opened 2018-07-24
DefaultLaunchCommandGeneratorTest according to the new template definitions

jvmafra:


src/test/java/org/fogbowcloud/manager/core/plugins/cloud/util/DefaultLaunchCommandGeneratorTest.java
```diff
@@ -135,105 +172,92 @@ public void testApplyTokensReplacements() {
...
         this.properties.setProperty(
-                ConfigurationConstants.REVERSE_TUNNEL_HTTP_PORT_KEY, this.reverseTunnelHttpPort);
+                ConfigurationConstants.REVERSE_TUNNEL_HTTP_PORT_KEY, REVERSE_TUNNEL_HTTP_PORT);
+        
+        // exercise
         new DefaultLaunchCommandGenerator();
     }
-
+    
+    // test case: An exception must be thrown because the reverse tunnel private address key is empty.
```
arn3tt: Isn't 'when' a better word than 'because' in this case?

src/test/java/org/fogbowcloud/manager/core/plugins/cloud/util/DefaultLaunchCommandGeneratorTest.java
```diff
@@ -135,105 +172,92 @@ public void testApplyTokensReplacements() {
...
-                ConfigurationConstants.REVERSE_TUNNEL_HTTP_PORT_KEY, this.reverseTunnelHttpPort);
+                ConfigurationConstants.REVERSE_TUNNEL_HTTP_PORT_KEY, REVERSE_TUNNEL_HTTP_PORT);
+        
+        // exercise
         new DefaultLaunchCommandGenerator();
     }
-
+    
+    
+    // test case: An exception must be thrown because the reverse tunnel http port key is empty.
```
arn3tt: Isn't 'when' a better word than 'because' in this case?

PR #119, opened 2018-07-26
refactor ClosedProcessor test code

thiagomanel:


src/test/java/org/fogbowcloud/manager/core/processors/ClosedProcessorTest.java
```diff
@@ -83,29 +76,39 @@ public void testProcessClosedLocalOrder() throws Exception {
         OrderStateTransitioner.activateOrder(localOrder);
 		OrderStateTransitioner.transition(localOrder, OrderState.CLOSED);
 
-        doNothing().when(this.localCloudConnector).deleteInstance(Mockito.any(Order.class));
+        Mockito.doNothing().when(this.localCloudConnector).deleteInstance(Mockito.any(Order.class));
```
arn3tt: Since `this.localCloudConnector` is a mock object, there's no need for this set up, right?

PR #204, opened 2019-02-22
Remove cache from CloudConnectorFactory and Create a cache of cloud p…

arn3tt:
…roperties in the plugins instantiator

### DIFF
--- a/src/main/java/cloud/fogbow/ras/core/InteroperabilityPluginInstantiator.java
+++ b/src/main/java/cloud/fogbow/ras/core/InteroperabilityPluginInstantiator.java
@@ -17,7 +17,7 @@ public class InteroperabilityPluginInstantiator {
     private PluginFactory pluginFactory;
     private Map<String, Properties> cloudPropertiesCache;
 
-    public InteroperabilityPluginInstantiator(String cloudName) {
+    public InteroperabilityPluginInstantiator() {
         this.pluginFactory = new PluginFactory();
         this.cloudPropertiesCache = new HashMap<>();
     }
--- a/src/main/java/cloud/fogbow/ras/core/InteroperabilityPluginInstantiator.java
+++ b/src/main/java/cloud/fogbow/ras/core/InteroperabilityPluginInstantiator.java
@@ -9,83 +9,93 @@ import cloud.fogbow.ras.core.plugins.interoperability.genericrequest.GenericRequ
 import cloud.fogbow.ras.core.plugins.mapper.FederationToLocalMapperPlugin;
 
 import java.io.File;
+import java.util.HashMap;
+import java.util.Map;
 import java.util.Properties;
 
 public class InteroperabilityPluginInstantiator {
-    private Properties properties;
-    private String cloudConfPath;
-    private String mapperConfPath;
     private PluginFactory pluginFactory;
+    private Map<String, Properties> cloudPropertiesCache;
 
     public InteroperabilityPluginInstantiator(String cloudName) {
+        this.pluginFactory = new PluginFactory();
+        this.cloudPropertiesCache = new HashMap<>();
+    }
+
+    private String getCloudProperty(String cloudName, String propertyKey) {
+        if (!cloudPropertiesCache.containsKey(cloudName)) {
+            Properties cloudProperties = readCloudProperties(cloudName);
+            cloudPropertiesCache.put(cloudName, cloudProperties);
+        }
+        return cloudPropertiesCache.get(cloudName).getProperty(propertyKey);
+    }
+
+    private String getCloudConfPath(String cloudName) {
         String path = HomeDir.getPath();
-        this.properties = PropertiesUtil.readProperties(path +
+        return path + SystemConstants.CLOUDS_CONFIGURATION_DIRECTORY_NAME + File.separator + cloudName +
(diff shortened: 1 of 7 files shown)


src/main/java/cloud/fogbow/ras/core/InteroperabilityPluginInstantiator.java
```diff
@@ -9,83 +9,93 @@
...
 import java.util.Properties;
 
 public class InteroperabilityPluginInstantiator {
-    private Properties properties;
-    private String cloudConfPath;
-    private String mapperConfPath;
     private PluginFactory pluginFactory;
+    private Map<String, Properties> cloudPropertiesCache;
 
     public InteroperabilityPluginInstantiator(String cloudName) {
```
fubica: We don't need a parameter for the constructor.

arn3tt: Nice catch, forgot to remove it.

src/main/java/cloud/fogbow/ras/core/cloudconnector/LocalCloudConnector.java
```diff
@@ -47,18 +47,20 @@
     private SecurityRulePlugin securityRulePlugin;
     private GenericRequestPlugin genericRequestPlugin;
 
+    private boolean shouldAuditRequests;
```
fubica: I suggest changing the name of this field to auditRequestOn.

src/main/java/cloud/fogbow/ras/core/cloudconnector/LocalCloudConnector.java
```diff
@@ -47,18 +47,20 @@
...
+
     public LocalCloudConnector(String cloudName) {
         InteroperabilityPluginInstantiator instantiator = new InteroperabilityPluginInstantiator(cloudName);
```
fubica: As per previous comment, there is no need to send cloudName in the constructor.

src/main/java/cloud/fogbow/ras/core/processors/FailedProcessor.java
```diff
@@ -86,7 +86,10 @@ protected void processFailedOrder(Order order) throws UnexpectedException {
             try {
                 LocalCloudConnector localCloudConnector = (LocalCloudConnector) CloudConnectorFactory.getInstance().
```
fubica: We don't need this casting, because it is guaranteed that the CloudConnector is a local one. Instead of this command, use:
LocalCloudConnector localCloudConnector = new LocalCloudConnector(cloudName);

src/main/java/cloud/fogbow/ras/core/processors/FulfilledProcessor.java
```diff
@@ -95,6 +95,10 @@ protected void processFulfilledOrder(Order order) throws UnexpectedException {
             try {
                 LocalCloudConnector localCloudConnector = (LocalCloudConnector) CloudConnectorFactory.getInstance().
```
fubica: We don't need this casting, because it is guaranteed that the CloudConnector is a local one. Instead of this command, use:
LocalCloudConnector localCloudConnector = new LocalCloudConnector(cloudName);

src/main/java/cloud/fogbow/ras/core/processors/SpawningProcessor.java
```diff
@@ -70,6 +70,9 @@ private void processInstance(Order order) throws FogbowException {
         LocalCloudConnector localCloudConnector = (LocalCloudConnector) CloudConnectorFactory.getInstance().
```
fubica: We don't need this casting, because it is guaranteed that the CloudConnector is a local one. Instead of this command, use:
LocalCloudConnector localCloudConnector = new LocalCloudConnector(order.getCloudName());

src/main/java/cloud/fogbow/ras/core/cloudconnector/LocalCloudConnector.java
arn3tt: Makes sense, renaming it.

src/main/java/cloud/fogbow/ras/core/cloudconnector/LocalCloudConnector.java
arn3tt: We do need the cloud name in this case because an instance of this class points to a specific cloud configured by the user.

src/main/java/cloud/fogbow/ras/core/processors/FailedProcessor.java
arn3tt: That's right, I will remove it.

src/main/java/cloud/fogbow/ras/core/processors/FulfilledProcessor.java
arn3tt: That's right, I will remove it.

src/main/java/cloud/fogbow/ras/core/cloudconnector/LocalCloudConnector.java
arn3tt: Right

PR #225, opened 2019-03-17
Truncate storable objects fields

arn3tt:


src/main/java/cloud/fogbow/ras/Main.java
```diff
@@ -93,6 +103,36 @@ public void run(ApplicationArguments args) {
         }
     }
 
+    private void getStorableObject(Object order) {
+        // new = constructor.call()
```
fubica: This code should be removed.

src/main/java/cloud/fogbow/ras/Main.java
```diff
@@ -93,6 +103,36 @@ public void run(ApplicationArguments args) {
...
+    private boolean isSizeViolation(TransactionSystemException e) {
+        Throwable e1 = e.getCause();
+        if (e1 != null && e1 instanceof RollbackException) {
+            Throwable e2 = e1.getCause();
+            if (e2 != null && e2 instanceof ConstraintViolationException) {
+                ConstraintViolationException constraintViolationException = (ConstraintViolationException) e2;
+                Set<ConstraintViolation<?>> constraintViolations = constraintViolationException.getConstraintViolations();
+                if (constraintViolations.iterator().hasNext()) {
+                    ConstraintViolation<?> constraintViolation = constraintViolations.iterator().next();
+                    if (constraintViolation.getMessage().startsWith("size must be between")) {
```
fubica: Is the message "size must be between" defined by us? If so, this should be defined as a constant in the appropriate class in the constant package, and properly documented in the places where it is used. If not, this should be defined as a constant in this class, and should also be properly documented.

src/main/java/cloud/fogbow/ras/core/datastore/services/FogbowDatabaseService.java
```diff
@@ -0,0 +1,77 @@
...
+import cloud.fogbow.ras.core.datastore.StorableObjectTruncateHelper;
+import org.springframework.data.jpa.repository.JpaRepository;
+import org.springframework.transaction.TransactionSystemException;
+
+import javax.persistence.RollbackException;
+import javax.validation.ConstraintViolation;
+import javax.validation.ConstraintViolationException;
+import java.util.Set;
+
+// TODO We might want to move this class to the Common project
```
fubica: Act on this TODO. The class should go to common if it is used by multiple services and if it does not import packages from any particular service. Currently this is not the case. Would it be possible to move cloud.fogbow.ras.core.datastore.StorableObjectTruncateHelper to common, as well? If not, then this class cannot be moved to common as it is.

src/main/java/cloud/fogbow/ras/core/models/orders/AttachmentOrder.java
```diff
@@ -82,14 +82,4 @@ public ResourceType getType() {
     public String getSpec() {
```
fubica: Please, add a TODO here and a task to solve this Spec stuff. This will be needed in the accounting service.

src/main/java/cloud/fogbow/ras/core/models/orders/ComputeOrder.java
```diff
@@ -41,7 +38,7 @@
     @Embedded
     private ArrayList<UserData> userData;
 
-    @Size(max = Order.FIELDS_MAX_SIZE)
+    @Size(max = 4)
```
fubica: I believe this size was fixed in 4 while debugging. Please have a look if this size should be fixed.

src/main/java/cloud/fogbow/ras/Main.java
arn3tt: Removing it...

src/main/java/cloud/fogbow/ras/Main.java
arn3tt: It's defined by Spring.

src/main/java/cloud/fogbow/ras/core/models/orders/ComputeOrder.java
arn3tt: Yes, setting it back.

src/main/java/cloud/fogbow/ras/core/datastore/services/FogbowDatabaseService.java
arn3tt: The only problem on moving this class to common is that we would need to add a few dependencies to common.

src/main/java/cloud/fogbow/ras/Main.java
arn3tt: Actually, this code is dead, I had forgotten to remove it. The actual code is in FogbowDatabaseService.isSizeViolation(). The message is defined by the JPA, so I created a constant field to hold its value.

src/main/java/cloud/fogbow/ras/core/models/orders/AttachmentOrder.java
arn3tt: Added.

PR #231, opened 2019-03-25
Dependencies check

pauloewerton:
Checks whether there's is a public ip or volume associated with a compute order before deleting it. Also, updates compute order dependencies list when a new attachment or public ip order is created.


fubica:
@pauloewerton please, have a look at the alternative implementation.

pauloewerton:
@fubica looks good, thanks. i still need to test this locally with the GUI before merging though.

fubica:
Great!

PR #236, opened 2019-03-27
Refactor structures

arn3tt:


fubica:
@arn3tt, please, have a look at the two commits I've added.

arn3tt:
lgtm

PR #240, opened 2019-03-28
Fix one network plugin

wassoares:
- Fixes made in the OpenNebulaNetworkPlugin class
- Fixed tests in the OpenNebulaNetworkPluginTest class and added more tests to increase coverage of this class.


fubica:
As I understood from the comments on the issues solved by this PR, this solution only works if the user that is used on the AllToOneMapper belongs to an admin group. We cannot make this kind of assumption. If a normal user is not able to create a network in opennebula, then we cannot provide this functionality for OpenNebula.

PR #241, opened 2019-03-29
Refact system user persistence

raonismaneoto:
The systemUser as a transient property in Order model was a big problem. The system needed the user saved to recovery from a possible crash, for example.
To solve this problem, and trying to avoid others, I've serialized the user with GSON and saved it before the order gets saved. On the other hand, when the order is retrieved, the user is reconstructed. For that, the private class SerializableSystemUser has been created to keep the information about what child class the user's instance was.

### DIFF
--- a/src/main/java/cloud/fogbow/ras/core/models/orders/Order.java
+++ b/src/main/java/cloud/fogbow/ras/core/models/orders/Order.java
@@ -2,10 +2,10 @@ package cloud.fogbow.ras.core.models.orders;
 
 import cloud.fogbow.common.exceptions.UnexpectedException;
 import cloud.fogbow.common.models.SystemUser;
+import cloud.fogbow.common.util.GsonHolder;
 import cloud.fogbow.ras.api.http.response.InstanceState;
 import cloud.fogbow.ras.core.datastore.DatabaseManager;
 import cloud.fogbow.ras.core.models.ResourceType;
-import com.google.gson.Gson;
 
 import javax.persistence.*;
 import javax.validation.constraints.Size;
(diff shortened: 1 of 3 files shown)


src/main/java/cloud/fogbow/ras/core/models/orders/Order.java
```diff
@@ -66,6 +67,18 @@
     @Transient
     private SystemUser systemUser;
 
+    @Column
+    @Size(max = FIELDS_MAX_SIZE)
+    private String userId;
+
+    @Column
+    @Size(max = 1024)
```
fubica: Will 1024 be enough? Please, have a look at the already defined SystemUser child classes to have an idea whether this size is adequate. Also, it would be good to add a comment there to warn developers that classes that extend SystemUser need to be aware of this size limit.

src/main/java/cloud/fogbow/ras/core/models/orders/Order.java
```diff
@@ -208,4 +262,19 @@ public String toString() {
...
     public abstract String getSpec();
+
+    private class SerializableSystemUser {
+
+        private String className;
+        private String payload;
+
+        public SerializableSystemUser(SystemUser user) {
+            this.className = user.getClass().getName();
+            this.payload = GsonHolder.getInstance().toJson(user);
```
fubica: Perhaps we should add a check here to see if the payload will fit in the database column (currently 1024 bytes - see previous comment), and throw an exception. Alternatively, we could do that at the SystemUser constructor. I am not sure, what would be the best choice, but this needs to be taken into account.

src/main/java/cloud/fogbow/ras/core/models/orders/Order.java
```diff
@@ -170,6 +183,47 @@ public void setRequirements(Map<String, String> requirements) {
...
+        this.setProviderId(this.getSystemUser().getIdentityProviderId());
+    }
+
+    @PostLoad
+    private void deserializeSystemUser() {
+        try {
+            SerializableSystemUser serializableSystemUser = GsonHolder.getInstance().fromJson(this.getSerializedSystemUser(), SerializableSystemUser.class);
+            this.setSystemUser(serializableSystemUser.getSystemUser());
+        } catch(ClassNotFoundException exception) {
+            exception.printStackTrace();
```
arn3tt: Standard output might be lost if the RAS process runs as a Daemon. It's better to have a Logger instance here and call Logger.error(). Take a look at ApplicationFacade to see how to do this.

PR #243, opened 2019-04-01
Remove orders when processing them on the ClosedProcessor instance

arn3tt:


src/main/java/cloud/fogbow/ras/core/processors/ClosedProcessor.java
```diff
@@ -46,9 +50,17 @@ public void run() {
...
         synchronized (order) {
-           this.orderController.deactivateOrder(order);
+            CloudConnector cloudConnector = this.orderController.getCloudConnector(order);
+            try {
+                cloudConnector.deleteInstance(order);
+            } catch (InstanceNotFoundException e) {
+                LOGGER.info(String.format(Messages.Info.DELETING_ORDER_INSTANCE_NOT_FOUND, order.getId()), e);
+            }
+
+            this.orderController.updateOrderDependencies(order, Operation.DELETE);
```
fubica: This should be part of the processing of the delete request. In fact, I now realize that perhaps we should really move all this code back to the delete processing, and do the same in the FNS.

PR #260, opened 2019-04-20
Remove cached instance logic

fubica:
This is a massive PR (sorry :-(, that fixed all the pending issues (and others found out on the way). It has been extensively tested in a local deployment. The main changes are the following:
- removed the clumsy logic of cached instance state that was kept in the order (now the instance state is inferred from the order state, and the instance object carries both this "fogbow" state, as well as the state that came from the cloud - which might be meaningful to the user; this required the creation of a new interface - OrderPlugin - for plugins that deal with orders)
- organized the logic that maps the order ids seen by the user, and the instance ids used internally (this mapping is now done solely at the OrderController class)
- refactored the logic for the orders  that have embedded orders (Attach, Compute and PublicIp)
- changed the logic for notifications of remote orders; previously, the local member was notified just once when the remote order moved for the first time to either failed or fulfilled - this was the source of the issue that would not update instance states for remote orders; now every time a remote order changes state (except when it goes to CLOSED) a notification is sent to the requesting member; this way, the local order and the corresponding remote order are kept in sync, regarding their state; a notification for the change to CLOSED is not needed because delete is a synchronous operation, so the local RAS can make the state change without relying on the remote RAS.


fubica:
In time ... this PR already integrates the changes in other branches related to ONE plugins.

src/main/resources/application.properties
```diff
@@ -9,7 +9,7 @@
 #spring.datasource.username=postgres
 #spring.datasource.password=postgres
 #spring.jpa.generate-ddl=true
-server.port=8081
+server.port=8080
```
raonismaneoto: I think this modification shouldn't be here

fubica: You're right in the sense that the value stored in the repository shouldn't be changed. This update is just to fix a change that was inadvertently made in a previous commit. This was probably due to someone (perhaps myself :-( testing several services in a local machine, in this case each service should run in a different port. However, in a real deployment the services will run in separate containers. Thus, we defined that all Fogbow services should listen on port 8080. The deployment procedure does the appropriate port mapping between the virtual machine port and the container that runs the service.

PR #269, opened 2019-05-08
Aws mapper impl

raonismaneoto:


src/main/java/cloud/fogbow/ras/core/plugins/mapper/all2one/AwsAllToOneMapper.java
```diff
@@ -0,0 +1,22 @@
+package cloud.fogbow.ras.core.plugins.mapper.all2one;
+
+import cloud.fogbow.common.exceptions.FogbowException;
+import cloud.fogbow.common.models.AwsUser;
+import cloud.fogbow.common.models.SystemUser;
+import cloud.fogbow.common.plugins.cloudidp.aws.AwsIdentityProviderPlugin;
+
+import java.util.Map;
+
+public class AwsAllToOneMapper extends GenericAllToOneSystemToCloudMapper<AwsUser, SystemUser> {
```
arn3tt: We might want to include the API version in this class name too.

PR #288, opened 2019-06-26
Check order params consistency

ANKerD:


src/main/java/cloud/fogbow/ras/api/parameters/Network.java
```diff
@@ -50,4 +58,12 @@ public String getCidr() {
     public NetworkAllocationMode getAllocationMode() {
         return allocationMode;
     }
+
+    private void checkCidrFormat(String cidr) throws FogbowException {
```
raonismaneoto: try to use common.Ipv4AddressValidator

src/main/java/cloud/fogbow/ras/api/parameters/SecurityRule.java
```diff
@@ -136,4 +139,10 @@ public String toString() {
             return this.protocol;
         }
     }
+
+    public void checkConsistency() throws FogbowException {
```
raonismaneoto: maybe you should check the cidr here too

PR #306, opened 2019-08-12
Fix recovery service tests

raonismaneoto:


src/test/resources/application.properties
```diff
@@ -0,0 +1,5 @@
+jdbc.driverClassName=org.h2.Driver
```
ANKerD: Add these conf to application-dev.

PR #329, opened 2019-09-12
Aws compute test refact

wassoares:
- Refactored the AwsV2ComputePlugin class;
- Added new constants to TestUtils class;
- Refactored all tests of the AwsV2ComputePluginTest class;
- Fixed code indent in these classes;
- And verify system tests in the AWS Cloud;


src/main/java/cloud/fogbow/ras/core/plugins/interoperability/aws/compute/v2/AwsV2ComputePlugin.java
```diff
@@ -48,430 +54,434 @@
...
+        return false;
+    }
+
+    protected void doDeleteInstance(String instanceId, Ec2Client client) throws UnexpectedException {
+        TerminateInstancesRequest request = TerminateInstancesRequest.builder()
+                .instanceIds(instanceId)
+                .build();
+        try {
+            client.terminateInstances(request);
+        } catch (Exception e) {
```
raonismaneoto: this is a sdkException, right?

raonismaneoto: check the same thing all over the methods please

wassoares: You're right, the exception to be caught on this line is a SdkException, but another occurrence on line 243, I believe it should not be changed because other exceptions can occur in the referenced block.

PR #375, opened 2019-12-17
New quota features

wassoares:
### Task Description
- Add new quota service to aggregate data to all cloud resource plug-ins.

### Proposed Changes
- Implement a new endpoint to access the total quota detached from the compute API.

### Additional information
- Unit tests will be added to the new-quota-feature-tests branch.

### DIFF
--- a/src/main/java/cloud/fogbow/ras/api/http/response/quotas/allocation/ResourceAllocation.java
+++ b/src/main/java/cloud/fogbow/ras/api/http/response/quotas/allocation/ResourceAllocation.java
@@ -1,0 +1,117 @@
+package cloud.fogbow.ras.api.http.response.quotas.allocation;
+
+import javax.persistence.Column;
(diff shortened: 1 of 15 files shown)


src/main/java/cloud/fogbow/ras/api/http/response/quotas/allocation/ResourceAllocation.java
```diff
@@ -0,0 +1,117 @@
+package cloud.fogbow.ras.api.http.response.quotas.allocation;
+
+import javax.persistence.Column;
```
pauloewerton: i'm unsure, but shouldn't we use the spring api?

wassoares: I used the same imports used in the **ComputeAllocation** class, I believe the Spring JPA implementation also works with these annotations.

PR #391, opened 2020-01-27
Allocation refactor integration

jadsonluan:


src/main/java/cloud/fogbow/ras/api/http/response/quotas/allocation/ComputeAllocation.java
```diff
@@ -34,7 +35,7 @@ public ComputeAllocation(int vCPU, int ram, int instances) {
         this.instances = instances;
     }
 
-    public ComputeAllocation() {
+    public ComputeAllocation(){
```
wassoares: Return the **space** between closing the parentheses and opening the keys.

PR #417, opened 2020-03-19
Azure volume plugin

wassoares:
This pull request implements the volume plugin for the Azure cloud.

### Additional information 
The unit tests were implemented in another branch.


src/main/java/cloud/fogbow/ras/constants/Messages.java
```diff
@@ -96,6 +96,7 @@
         public static final String CONNECTING_UP_PACKET_SENDER = "Connecting XMPP packet sender.";
         public static final String DELETING_INSTANCE = "Deleting instance %s with token %s.";
         public static final String DELETING_INSTANCE_S = "Deleting instance %s.";
+        public static final String END_CREATE_DISK_ASYNC_BEHAVIOUR = "End asynchronous create disk";
```
fgan1: Add "." at the end of the text.

src/main/java/cloud/fogbow/ras/constants/Messages.java
```diff
@@ -140,6 +141,7 @@
     public static class Error {
     	public static final String CONTENT_SECURITY_GROUP_NOT_DEFINED = "The content of SecuriryGroups in the VirtualNetwork template is not defined.";
         public static final String COULD_NOT_FIND_DEPENDENCY_S_S = "Could not find dependency %s for order %s.";
+        public static final String ERROR_CREATE_DISK_ASYNC_BEHAVIOUR = "Error while creating disk asynchronously";
```
fgan1: Add "." at the end of the text.

src/main/java/cloud/fogbow/ras/core/plugins/interoperability/azure/volume/AzureVolumePlugin.java
```diff
@@ -0,0 +1,158 @@
...
+        return AzureStateMapper.map(ResourceType.VOLUME, instanceState).equals(InstanceState.READY);
+    }
+
+    @Override
+    public boolean hasFailed(String instanceState) {
+        return AzureStateMapper.map(ResourceType.VOLUME, instanceState).equals(InstanceState.FAILED);
+    }
+
+    @Override
+    public String requestInstance(VolumeOrder volumeOrder, AzureUser azureUser) throws FogbowException {
```
fgan1: How long takes the Azure Disk operation in the cloud? Is it possible to cause the False Failed Order problem? Do you remember the bug? Talk with me in the Mattermost.

wassoares: The volume or simply _"disk"_ in the context of the Azure cloud, is one of the features that take less time to get ready. But I believe that the asynchronous plugin strategy is still the best even in this case. In system tests carried out in the cloud I did not have any problem of order failure, using either the local region _"brazilsouth"_ or the default _"eastus"_.

src/main/java/cloud/fogbow/ras/constants/Messages.java
wassoares: I'll fix it.

src/main/java/cloud/fogbow/ras/constants/Messages.java
wassoares: I'll fix it.

PR #423, opened 2020-04-01
Fix azure compute plugin

wassoares:
## Description
This pull request fixes the network interface connection between compute and network resources when creating the Virtual Machine in the Azure cloud.

## Dependencies
This branch also depends on the acceptance of the branch of the same name in the common project.


src/main/java/cloud/fogbow/ras/constants/Messages.java
```diff
@@ -97,6 +97,7 @@
         public static final String CONNECTING_UP_PACKET_SENDER = "Connecting XMPP packet sender.";
         public static final String DELETING_INSTANCE = "Deleting instance %s with token %s.";
         public static final String DELETING_INSTANCE_S = "Deleting instance %s.";
+        public static final String END_DELETE_NIC_ASYNC_BEHAVIOUR = "End asynchronous delete network interface";
```
fgan1, line 100: Add the "." at the end of the text in order to follow the pattern.

wassoares: Fixed.

PR #426, opened 2020-04-22
Azure publicip plugin

wassoares:
This pull request implements the public IP plugin for the Azure cloud.

### Additional information
The unit tests were implemented in another branch.


src/main/java/cloud/fogbow/ras/core/plugins/interoperability/azure/publicip/sdk/AzurePublicIPAddressOperationSDK.java
```diff
@@ -0,0 +1,133 @@
...
+    @VisibleForTesting
+    Observable<NetworkInterface> setAssociatePublicIPAddressBehaviour(Azure azure, String instanceId,
+            Observable<NetworkInterface> observable) {
+
+        return observable.doOnNext(nic -> {
+            LOGGER.info(Messages.Info.FIRST_STEP_CREATE_PUBLIC_IP_ASYNC_BEHAVIOUR);
+            handleSecurityIssues(azure, instanceId, nic);
+            LOGGER.info(Messages.Info.SECOND_STEP_CREATE_AND_ATTACH_NSG_ASYNC_BEHAVIOUR);
+        }).onErrorReturn(error -> {
+            LOGGER.error(Messages.Error.ERROR_CREATE_PUBLIC_IP_ASYNC_BEHAVIOUR, error);
```
fgan1, line 55: In this point. Should't you remove the Public Ip if the Security Group Creation fail?

src/main/java/cloud/fogbow/ras/core/plugins/interoperability/azure/publicip/sdk/AzurePublicIPAddressOperationSDK.java
```diff
@@ -0,0 +1,133 @@
...
+            .subscribe();
+    }
+
+    @VisibleForTesting
+    Observable<NetworkInterface> setAssociatePublicIPAddressBehaviour(Azure azure, String instanceId,
+            Observable<NetworkInterface> observable) {
+
+        return observable.doOnNext(nic -> {
+            LOGGER.info(Messages.Info.FIRST_STEP_CREATE_PUBLIC_IP_ASYNC_BEHAVIOUR);
+            handleSecurityIssues(azure, instanceId, nic);
```
fgan1, line 52: I would use another method name. This is an important step in our operation, therefore we should make it explicit.
We might use associateSecurityGroup. What do you think?
Or you might use the same way of other plugins. Something like: [Link](https://github.com/fogbow/resource-allocation-service/blob/develop/src/main/java/cloud/fogbow/ras/core/plugins/interoperability/azure/network/sdk/AzureVirtualNetworkOperationSDK.java#L70)

wassoares: Ok, I'll rename this method.

src/main/java/cloud/fogbow/ras/core/plugins/interoperability/azure/publicip/sdk/AzurePublicIPAddressOperationSDK.java
wassoares: Ok, I'll implement a method to remove the public IP if it was created.

PR #434, opened 2020-05-04
Update azure volume plugin

wassoares:
This pull request implements the changes made to the Azure volume plugin for the deletion of resources via the resource group.

## Reminding
Pull requests order:
azure-resource-group-integration < update-azure-volume-plugin


src/main/java/cloud/fogbow/ras/core/plugins/interoperability/azure/volume/AzureVolumePlugin.java
```diff
@@ -152,6 +177,17 @@ void updateInstanceAllocation(VolumeOrder volumeOrder) {
         }
     }
     
+    @VisibleForTesting
+    String defineResourceGroupName(Azure azure, String resourceName) {
+        String resourceGroupName = this.defaultResourceGroupName;
+        try {
+            resourceGroupName = AzureResourceGroupUtil.create(azure, this.defaultRegionName, resourceName);
+        } catch (Exception e) {
+            LOGGER.error(String.format(Messages.Error.ERROR_MESSAGE, e));
```
fgan1, line 186: Use "info" or "warn" instead of "error" in the log level. Moreover, you could specify a better error message because this is an important step in code logic.

src/main/java/cloud/fogbow/ras/core/plugins/interoperability/azure/volume/AzureVolumePlugin.java
```diff
@@ -152,6 +177,17 @@ void updateInstanceAllocation(VolumeOrder volumeOrder) {
         }
     }
     
+    @VisibleForTesting
+    String defineResourceGroupName(Azure azure, String resourceName) {
```
fgan1, line 181: What do you think about this suggestion ?
```suggestion
    @VisibleForTesting
    String defineResourceGroupName(Azure azure, String resourceName) {
        try {
            resourceGroupName = AzureResourceGroupUtil.create(azure, this.defaultRegionName, resourceName);
        } catch (Exception e) {
            LOGGER.debug("", e)        
            LOGGER.info("Something the explain this normal and predicted situation"));
            return this.defaultResourceGroupName;
        }
    }
```

src/main/java/cloud/fogbow/ras/core/plugins/interoperability/azure/volume/AzureVolumePlugin.java
```diff
@@ -94,10 +100,14 @@ public void deleteInstance(VolumeOrder volumeOrder, AzureUser azureUser) throws
         LOGGER.info(String.format(Messages.Info.DELETING_INSTANCE_S, volumeOrder.getInstanceId()));
         Azure azure = AzureClientCacheManager.getAzure(azureUser);
         String resourceName = AzureGeneralUtil.defineResourceName(volumeOrder.getInstanceId());
-        String subscriptionId = azureUser.getSubscriptionId();
-        String resourceId = buildResourceId(subscriptionId, resourceName);
         
-        doDeleteInstance(azure, resourceId); 
+        if (AzureResourceGroupOperationUtil.existsResourceGroup(azure, resourceName)) {
+            doDeleteResourceGroup(azure, resourceName);
```
fgan1, line 105: Will this method repeat in other plugins? Maybe you can use an auxiliary class.

src/main/java/cloud/fogbow/ras/core/plugins/interoperability/azure/volume/AzureVolumePlugin.java
```diff
@@ -83,9 +87,11 @@ public VolumeInstance getInstance(VolumeOrder volumeOrder, AzureUser azureUser)
         LOGGER.info(String.format(Messages.Info.GETTING_INSTANCE_S, volumeOrder.getInstanceId()));
         Azure azure = AzureClientCacheManager.getAzure(azureUser);
         String resourceName = AzureGeneralUtil.defineResourceName(volumeOrder.getInstanceId());
+        String resourceGroupName = AzureGeneralUtil
+                .selectResourceGroupName(azure, resourceName, this.defaultResourceGroupName);
```
fgan1, line 91: Is there an option to avoid to do another request to Azure cloud. 
If you jsut check the resourceName and the resourceNameGroup?

wassoares: Unfortunately, I not found another option.

src/main/java/cloud/fogbow/ras/core/plugins/interoperability/azure/volume/AzureVolumePlugin.java
wassoares: This point is decisive in which option should be followed at the time of exclusion, the first option is obvious to remove via resource group but otherwise, it will always change according to each plugin. If you have a better suggestion please share.

Issue #229, opened 2019-03-20
[OpenStack] Get Public IP order security rules returns null CIDR on all rules


pauloewerton:

fubica:
In fact, it should have four rules, as follows:

Direction | Ether Type | IP Protocol | Port Range | CIDR
Egress       IPv4              Any              Any               0.0.0.0/0
Egress       IPv6              Any              Any               ::/0
Ingress      IPv4              TCP             22                 0.0.0.0/0
Ingress      IPv6              TCP             22                 ::/0

pauloewerton:
@fubica thanks for the correction i will update the title. also, in my tests, the security rules listing from a public ip order from cloudstack returns only the ipv4 tcp 22 rule.

fubica:
Here the most important thing is to guarantee a common semantics for Fogbow, independent from the underlying cloud. The idea is that when a public IP is attached to a compute, then port 22 is open to anyone willing to connect to the VM (how this is implemented by the underlying cloud can vary).

pauloewerton:
we probably should be using "remote_ip_prefix" as the key to get cidr in our json response object. refer to: https://developer.openstack.org/api-ref/network/v2/index.html?expanded=list-floating-ips-detail,show-security-group-rule-detail#show-security-group-rule

fubica:
I believe that CIDR is the correct way to call this field (see https://pt.wikipedia.org/wiki/Classless_Inter-Domain_Routing). I think OpenStack folks got that one wrong. See for example AWS syntax (https://medium.com/@dbclin/a-bash-script-to-update-an-existing-aws-security-group-with-my-new-public-ip-address-d0c965d67f28).

pauloewerton:
@fubica yes, i misread our code. we're already using gson to map "remote_ip_prefix" to cidr in our json response object. please, ignore my previous comment.

pauloewerton:
we're not specifically passing a cidr (remote ip prefix) when creating a public ip default security rules. please check: https://github.com/fogbow/resource-allocation-service/blob/develop/src/main/java/cloud/fogbow/ras/core/plugins/interoperability/openstack/publicip/v2/OpenStackPublicIpPlugin.java#L217. in the openstack api, that parameter is optional so it will return null on get operation, meaning "allow to all". 

we can create a filter on the clients, such as the GUI, to present that data differently. horizon values for this case are "-" (ingress), "0.0.0.0/0" and "::/0" (egress).

[...]

PR #477, opened 2020-06-04
Fix openstack securityrule plugin

wassoares:
## Description
This pull request implements the changes suggested in **Issue** #229 of the resource allocation service project, to correct the null value return in the CIDR field in security rules.

## Proposed changes
- Fixed the construction of the create security group request by adding port 22 for SSH access, in the _allowAllIngressSecurityRules_ method of the **OpenStackPublicIpPlugin** class;
- Changes made in some attribute types from the **GetSecurityRulesResponse** class;
- And fixes made in the **OpenStackSecurityRulePlugin** classes, with the refactoring of methods and increased test coverage.

## Dependencies
This request depends on the branch of the same name (**fix-openstack-securityrule-plugin**) in the **common** project.


src/test/java/cloud/fogbow/ras/core/plugins/interoperability/openstack/securityrule/v2/OpenStackSecurityRulesPluginTest.java
```diff
@@ -43,474 +51,844 @@
...
-    
-    private static final String ANY_STRING = "any-string";
-    private static final String ANY_URL = "http://localhost:8007";
-    private static final String NETWORK_NEUTRONV2_URL_KEY = "openstack_neutron_v2_url";
-    private static final String DEFAULT_NETWORK_URL = "http://localhost:0000";
-    private static final String INGRESS_TRAFFIC = "ingress";
-    private static final String SECURITY_RULE_ID = "securityRuleId";
-    private static final String SECURITY_GROUP_ID = "securityGroupId";
+    private static final String NETWORK_PREFIX_ENDPOINT = "https://mycloud.domain:9696";
+    private static final String QUERY_SECURITY_GROUP_ID = "?security_group_id=";
```
fgan1, line 55: There are ways more elegant to create this "query" strings and JSON but I will not ask to change it now because we need to review all plugins to do it.

src/main/java/cloud/fogbow/ras/core/plugins/interoperability/openstack/securityrule/v2/OpenStackSecurityRulePlugin.java
```diff
@@ -3,110 +3,113 @@
...
 import java.util.ArrayList;
 import java.util.List;
 import java.util.Properties;
 
 public class OpenStackSecurityRulePlugin implements SecurityRulePlugin<OpenStackV3User> {
+
     private static final Logger LOGGER = Logger.getLogger(OpenStackSecurityRulePlugin.class);
+    private static final int MAXIMUM_SIZE_ALLOWED = 1;
+
+    public static final int MAXIMUM_PORT_RANGE = 65535;
```
fgan1, line 44: Are we using these constants in other plugins? If yes, you could refactor to a unified class.

src/test/java/cloud/fogbow/ras/core/plugins/interoperability/openstack/securityrule/v2/OpenStackSecurityRulesPluginTest.java
wassoares: Certainly, I really liked the way you did the tests on CloudStack, but here it was not practical to perform this kind of test refactoring on this task.

src/main/java/cloud/fogbow/ras/core/plugins/interoperability/openstack/securityrule/v2/OpenStackSecurityRulePlugin.java
wassoares: Not exactly, there is a similar constant **DEFAULT_RANGE_PORTS** in the **AzureNetworkSecurityGroupOperationSDK** class, but its value represents the range from _0_ to _65535_, not the maximum value required here.

PR #485, opened 2020-06-09
Fix azure failed on request context

### LINKED ISSUE #473: [Azure] Get instance operation fails when there an error in the SDK creation request.
Since Azure plugin has an asynchronous creation request, when the request fails and does not create an instance in the Azure cloud, the Fogbow core is cheated due to the fact that it wants for an instance.
Thus, if there is no instance in the cloud and there will be a getInstance operation it will throw an exception and it should be an instance failed.

Fogbow Flow Expected
requestInstance --->  Error
getInstance ---> Instance Failed

Current Fogbow Flow 
requestInstance --->  Error
getInstance ---> throw an exception

fgan1:
## Description
fixes #473 

## Proposed changes
- Azure Plugins deal with FAILED_ON_REQUEST state.
It must wait some time (1 second) in order to catch Azure SDK errors and throw an exception when it is necessary.
- Azure Plugins deal with FAILED_AFTER_SUCCESSFUL_REQUEST state.

## Reminder
PRs Orders: 
develop <-- this


src/test/java/cloud/fogbow/ras/core/plugins/interoperability/azure/compute/AzureComputePluginTest.java
```diff
@@ -222,6 +212,9 @@ public void testRequestInstanceSuccessfully() throws Exception {
         Mockito.doReturn(virtualMachineSize).when(this.azureComputePlugin)
                 .getVirtualMachineSize(Mockito.eq(computeOrder), Mockito.eq(this.azureUser));
 
+        Mockito.doReturn("").when(this.azureComputePlugin)
```
wassoares, line 215: You can use the EMPTY_STRING constant of the TestUtil class here.

fgan1: done

PR #707, opened 2020-10-22
Implemented Google Cloud Compute Plugin

jeron7:
## Description
Added the GoogleCloudCompute Plugin. Left two methods (isReady and hasFailed) to be implemented.

## Proposed changes
Build the missing methods and make some changes in requestInstance (doubt about Security Rules)

### DIFF
--- a/src/main/java/cloud/fogbow/ras/core/plugins/interoperability/googlecloud/compute/v1/GoogleCloudComputePlugin.java
+++ b/src/main/java/cloud/fogbow/ras/core/plugins/interoperability/googlecloud/compute/v1/GoogleCloudComputePlugin.java
@@ -118,7 +118,8 @@ public class GoogleCloudComputePlugin implements ComputePlugin<GoogleCloudUser> 
     private ComputeInstance doGetInstance(String endpoint, GoogleCloudUser cloudUser) throws FogbowException {
         String responseJson = this.client.doGetRequest(endpoint, cloudUser);
         ComputeInstance computeInstance = getInstanceFromJson(responseJson);
-        return null;
+
+        return computeInstance;
     }
 
     private void doDeleteRequest(String endpoint, GoogleCloudUser cloudUser) throws FogbowException {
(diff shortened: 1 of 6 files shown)


src/main/java/cloud/fogbow/ras/core/plugins/interoperability/googlecloud/compute/v1/GoogleCloudComputePlugin.java
```diff
@@ -0,0 +1,275 @@
+package cloud.fogbow.ras.core.plugins.interoperability.googlecloud.compute.v1;
+
+import cloud.fogbow.common.exceptions.FatalErrorException;
+import cloud.fogbow.common.exceptions.FogbowException;
+import cloud.fogbow.common.exceptions.InternalServerErrorException;
+import cloud.fogbow.common.models.GoogleCloudUser;
+import cloud.fogbow.common.util.PropertiesUtil;
+import cloud.fogbow.ras.core.plugins.interoperability.googlecloud.sdk.v1.compute.models.CreateComputeRequest;
+import cloud.fogbow.ras.core.plugins.interoperability.googlecloud.sdk.v1.compute.models.CreateComputeResponse;
+import static cloud.fogbow.ras.core.plugins.interoperability.googlecloud.sdk.v1.compute.models.CreateComputeRequest.*;
```
fubica, line 10: Better use import, instead of import static, since this will force you to explicitly refer to the static class name in the code, making it clearer that you are invoking a method from a static class. This leaves the code more readable and maintainable.

src/main/java/cloud/fogbow/ras/core/plugins/interoperability/googlecloud/compute/v1/GoogleCloudComputePlugin.java
```diff
@@ -0,0 +1,275 @@
...
+import static cloud.fogbow.ras.core.plugins.interoperability.googlecloud.util.GoogleCloudConstants.*;
```
fubica, line 11: Same as above.

fubica (changes requested):
Good job!

PR #746, opened 2020-11-21
Delete attachment fixed

quirinoflavio:
## Description
The method used to delete an attachment is POST.

### DIFF
--- a/src/main/java/cloud/fogbow/ras/core/plugins/interoperability/googlecloud/attachment/v1/GoogleCloudAttachmentPlugin.java
+++ b/src/main/java/cloud/fogbow/ras/core/plugins/interoperability/googlecloud/attachment/v1/GoogleCloudAttachmentPlugin.java
@@ -30,17 +30,15 @@ import java.util.Properties;
 public class GoogleCloudAttachmentPlugin implements AttachmentPlugin<GoogleCloudUser> {
     private static final Logger LOGGER = Logger.getLogger(GoogleCloudAttachmentPlugin.class);
 
-
-    //To delete an instance the body must be empty.
-    //Somehow content_length is not added to headers when the request body is empty.
-    //So the request will return a code > 204 (HttpStatus.SC_NO_CONTENT) and CloudHttpClient will throw an exception.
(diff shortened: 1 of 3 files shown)


src/main/java/cloud/fogbow/ras/core/plugins/interoperability/googlecloud/attachment/v1/GoogleCloudAttachmentPlugin.java
```diff
@@ -117,7 +118,7 @@ AttachmentInstance buildAttachmentInstanceFrom(GetAttachmentResponse response) {
         String device = response.getDevice();
 
         // TODO - Can I use 'status' from Cloud response?
-        String googleCloudState = EMPTY_STRING;
+        String googleCloudState = GoogleCloudConstants.EMPTY_STRING;
```
fubica, line 121: Not only you can, but you should. This field is called cloudState. It should not be confused with OrderState.

src/main/java/cloud/fogbow/ras/core/plugins/interoperability/googlecloud/attachment/v1/GoogleCloudAttachmentPlugin.java
```diff
@@ -30,8 +30,11 @@
 public class GoogleCloudAttachmentPlugin implements AttachmentPlugin<GoogleCloudUser> {
     private static final Logger LOGGER = Logger.getLogger(GoogleCloudAttachmentPlugin.class);
 
-    @VisibleForTesting
-    static final String EMPTY_STRING = "";
+
+    //To delete an instance the body must be empty.
+    //Somehow content_length is not added to headers when the request body is empty.
+    //So the request will return a code > 204 (HttpStatus.SC_NO_CONTENT) and CloudHttpClient will throw an exception.
+    static final String SPECIAL_EMPTY_BODY = "{null:null}";
```
fubica, line 37: Does the above comment still make sense? If not, remove. If yes, then perhaps you might want to rephrase it to make it clearer.

fubica (changes requested):

PR #754, opened 2020-12-01
Fixed ComposedAuthorizationPlugin cast problem

armstrongmsg:
## Description
This PR fixes the issue on using ComposedAuthorizationPlugin in RAS. Normally, using this class as authorization plugin would cause an ClassCastException to be thrown.

## Proposed changes
Pass a ClassFactory created by RAS to the ComposedAuthorizationPlugin instance. This classFactory should be the one used to create the sub-plugins instances.

## Additional information
Requires https://github.com/fogbow/common/pull/271

## Reminding
PRs order:
current < your(PR1) < your(PR2)

### DIFF
--- a/src/main/java/cloud/fogbow/ras/core/RasClassFactory.java
+++ b/src/main/java/cloud/fogbow/ras/core/RasClassFactory.java
@@ -1,15 +1,18 @@
 package cloud.fogbow.ras.core;
 
 import cloud.fogbow.common.exceptions.FatalErrorException;
+import cloud.fogbow.common.plugins.authorization.ComposedAuthorizationPlugin;
+import cloud.fogbow.common.util.ClassFactory;
 import cloud.fogbow.ras.constants.Messages;
+import cloud.fogbow.ras.core.models.RasOperation;
 
 import java.lang.reflect.Constructor;
 import java.util.Arrays;
 
 // Each package has to have its own ClassFactory
 
-public class ClassFactory {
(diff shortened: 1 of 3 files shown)


src/main/java/cloud/fogbow/ras/core/RasClassFactory.java
```diff
@@ -27,6 +30,11 @@ public Object createPluginInstance(String pluginClassName, String ... params)
             }
 
             pluginInstance = constructor.newInstance(params);
+            
+        	if (pluginClassName.equals("cloud.fogbow.common.plugins.authorization.ComposedAuthorizationPlugin")) {
+        		ComposedAuthorizationPlugin<RasOperation> composedPlugin = (ComposedAuthorizationPlugin<RasOperation>) pluginInstance;
+        		composedPlugin.startPlugin(this);
+        	}
```
fubica, line 37: Smells bad. At a minimum I would move this code to the getAuthorizationPlugin() method in AuthorizationPluginInstantiator.

fubica (changes requested):

src/main/java/cloud/fogbow/ras/core/RasClassFactory.java
armstrongmsg: Moved to AuthorizationPluginInstantiator.

fubica (approved):

